PR #204, opened 2023-02-11
Sidebar fixes

mark-tate:


packages/schemas/src/sidebarDataLayoutSchema.ts
```diff
@@ -1,9 +1,16 @@
 import { z } from 'zod';
 
 export const sidebarDataLayoutSchema = z.union([
```
DavieReid, line 3: Only the layouts that had a sidebar were added here...why would we need to add layouts that don't have a sidebar?

DavieReid (approved):

packages/schemas/src/sidebarDataLayoutSchema.ts
mark-tate: Good shout... I had a sidebar and changed the layout and then saw the error and thought this schema validated layouts.. was on my radar to ask you, if folks add their own layouts, this would need to be extended.
I can put it back to what it was (my bad). Although, how do we see folks adding their own layouts that use sidebars ?

mark-tate: restored previous change

PR #215, opened 2023-02-16
Sidebar improvements

mark-tate:


packages/plugins/src/SidebarPlugin.ts
```diff
@@ -4,7 +4,7 @@ import { sidebarDataLayoutSchema } from '@jpmorganchase/mosaic-schemas';
 import { cloneDeep } from 'lodash-es';
 
 // What level the sidebar.json files are created
-const UserJourneyRootLevel = 2;
+const UserJourneyRootLevel = 1;
```
lilyvc, line 7: Could this be an option in the future

lilyvc (approved):

PR #216, opened 2023-02-16
feat: hide login button when not logging in is disabled

DavieReid:
NOTE: env variable has `NEXT_PUBLIC` prefix so that the value is available client-side.

### DIFF
--- a/packages/site-components/src/AppHeaderControls/index.tsx
+++ b/packages/site-components/src/AppHeaderControls/index.tsx
@@ -29,6 +29,8 @@ export const AppHeaderControls: React.FC<HeaderControlsProps> = () => {
   const { setColorMode } = useStoreActions();
 
   const { data: session } = useSession();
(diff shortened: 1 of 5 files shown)


tomhazledine (approved):
Nice. LGTM

Is `ENABLE_LOGIN` totally superseded by `NEXT_PUBLIC_ENABLE_LOGIN`?

And is `withSession` still in the codebase? IIRC it was not used, but still in the repo. That used `ENABLE_LOGIN` too I think...

DavieReid:
> Nice. LGTM
> 
> Is `ENABLE_LOGIN` totally superseded by `NEXT_PUBLIC_ENABLE_LOGIN`?
> 
> And is `withSession` still in the codebase? IIRC it was not used, but still in the repo. That used `ENABLE_LOGIN` too I think...

It will supersede yeah.  We need to use the value on the client so if we don't add the prefix we get a hydration error.  good point about withSession, let me update that 1

PR #217, opened 2023-02-16
feat: default to `en` locale

DavieReid:


tomhazledine (approved):
Oh gosh, does this mean that proper i18n is on the horizon for us?
danger

origami-z:
How do i do translation 😸

DavieReid:
> Oh gosh, does this mean that proper i18n is on the horizon for us? danger

I just need the html element to have a lang attribute at this point lol

PR #219, opened 2023-02-17
Sidebar improvements

lukeac123:


packages/site-components/src/VerticalNavigation.tsx
```diff
@@ -17,11 +17,26 @@ export type VerticalNavigationProps = {
 const MenuIcon = ({ open }) => <Icon name={open ? 'chevronDown' : 'chevronRight'} />;
 
 const menuItemStyles = {
-  button: ({ active }) => {
-    return {
-      fontWeight: active ? 'var(--fontWeight-bold)' : 'var(--fontWeight-regular)'
-    };
+  root: {
+    fontSize: '13px'
```
lukeac123, line 21: needs to be plugged into vanilla extract theme

packages/site-components/src/VerticalNavigation.tsx
```diff
@@ -17,11 +17,26 @@ export type VerticalNavigationProps = {
...
+  },
+  subMenuContent: {
+    backgroundColor: 'inherit'
+  },
+  button: ({ active }) => ({
+    fontWeight: active ? 'var(--fontWeight-bold)' : 'var(--fontWeight-regular)',
+    '&:hover': {
+      opacity: 0.5,
```
lukeac123, line 29: temporary fix - would useColorMode() should be used to switch on hover colours between light and dark modes but not working in this file for some reason

packages/theme/src/sidebar/sidebar.css.ts
```diff
@@ -24,8 +24,9 @@ export const sidebar = recipe({
       bottom: 0,
       overflowY: 'scroll',
       position: 'relative',
-      top: `${config.appHeader.height}px`,
-      height: `calc(100% - ${config.appHeader.height}px)`
+      top: '-20px'
```
lukeac123, line 27: temporary fix, work needs to be done on layouts for positioning of sidebar

packages/site-components/src/VerticalNavigation.tsx
mark-tate: i think Fusion asked for it to be the same size as the content.. i removed  the action1 style we added previously.. if we need it back to 13px, have a look at the lines I removed and we can re-add/refactor as needed

mark-tate:
looks like this PR has my old  commits on it..
Probably need to re-fetch to sync your git index.

Also worth looking at the other PR I pushed as it fixes up a few issues and give you some examples of how to use the theme to configure salt variables.
Hope it helps.

(Feel free to change)

PR #221, opened 2023-02-20
sidebar style fixes

lukeac123:


packages/site-components/src/VerticalNavigation.tsx
```diff
@@ -54,7 +54,8 @@ const createMenuItemStyles = colorMode => ({
 });
 
 const rootStyles = {
-  borderRightWidth: '0px'
+  borderRightWidth: '1px',
```
mark-tate, line 57: the problem with the border is that it exposes a new issue, that our footer is not full width.. we need to refactor layouts, if design folks want a full width footer?

That might not be something we want to do right now.

You do not need to set this as it is the default, the previous edit, turned it off, remove line 57 and it will be back.

Another new issue of adding the border is that it's the pro-sidebar default grey, so we would need to add the variables like we have done for the highlight color. The border changes color between light and dark mode.

Our theme has variables for border colors, so we could use those.

packages/site-components/src/VerticalNavigation.tsx
```diff
@@ -66,6 +67,8 @@ const SubMenuLink = ({ href, selectedNodeId, ...rest }) => {
 
 const renderMenu = (menu, expandedNodeIds, selectedNodeId) =>
   menu.reduce((result, item) => {
+    console.log(result);
```
mark-tate, line 70: worth removing

packages/theme/src/sidebar/sidebar.css.ts
```diff
@@ -1,8 +1,9 @@
 import { style } from '@vanilla-extract/css';
 import { defineProperties, createSprinkles } from '@vanilla-extract/sprinkles';
 import { recipe, RecipeVariants } from '@vanilla-extract/recipes';
+// import { globalStyle } from '@vanilla-extract/css';
```
mark-tate, line 4: worth removing

packages/theme/src/sidebar/sidebar.css.ts
```diff
@@ -25,8 +26,9 @@ export const sidebar = recipe({
       display: 'flex',
       overflowY: 'scroll',
       position: 'relative',
-      top: `${config.appHeader.height}px`,
-      height: 'auto'
+      top: '-20px',
+      // top: `${config.appHeader.height}px`,
+      height: '100%'
```
mark-tate, line 31: -20px is a magic number, would look at what we can use from the theme.

As before, not covering the app header, creates new issues, as the close button appears twice. Would ask design folks to help us work out the best solution.

PR #222, opened 2023-02-20
Remove patch-package from site workspace

joshwooding:


joshwooding:
> Is this only causing an issue because we use a monorepo?
> 
> Because if it's just a general problem, then we would also need to remove this from the standard generator as it currently creates a patches folder and adds in the commands you have removed.

I believe it's due to it being in a monorepo and using yarn v1 and patch-package. When I tested yarn v3 it didn't have the same issue since the patching mechanism is quite different.

PR #223, opened 2023-02-20
Fix Windows compatibility

joshwooding:


packages/core/src/plugin/loadDefinitionModules.ts
```diff
@@ -44,7 +47,7 @@ export default async function loadDefinitionModules(
             | Partial<Plugin | Serialiser>
             | { __esModule: boolean; default: Partial<Plugin | Serialiser> };
           // eslint-disable-next-line no-await-in-loop
-        } = await import(modulePath);
+        } = await import(pathToFileURL(require.resolve(modulePath)).toString());
```
mark-tate, line 50: will look into later but could not merge this as 1 test failed for me

  ● GIVEN Source › WHEN calling `use` › THEN the plugins should be passed to the worker

    Could not load plugin/serialiser modules due to error.
    Error: Cannot find module 'plugin' from 'packages/core/src/plugin/loadDefinitionModules.ts'

      63 |     return results;
      64 |   } catch (e) {
    > 65 |     throw new Error(`Could not load plugin/serialiser modules due to error.
         |           ^
      66 | ${e}`);
      67 |   }
      68 | }

      at loadDefinitionModules (packages/core/src/plugin/loadDefinitionModules.ts:65:11)
      at createPluginAPI (packages/core/src/plugin/createPluginAPI.ts:48:89)
      at bindPluginMethods (packages/core/src/plugin/index.ts:39:38)
      at Source.start (packages/core/src/Source.ts:180:46)
      at Object.<anonymous> (packages/core/src/__tests__/Source.test.ts:123:7)

mark-tate (approved):

PR #226, opened 2023-02-22
Hiding searchbar on small screens

tomhazledine:
Reattaches "hide on mobile" behaviour for the search input in the header.


mark-tate:
Search works nicely and search now appears in mobile as intended.

but when in mobile it truncates the menu

Screenshot 2023-02-22 at 12 01 00

mark-tate (approved):

mark-tate:
ah we are hiding it on mobile .. gotcha

PR #228, opened 2023-02-22
Sidebar priority

lukeac123:
Resolve issues with priority and fix sidebar width

### DIFF
--- a/packages/plugins/src/SidebarPlugin.ts
+++ b/packages/plugins/src/SidebarPlugin.ts
@@ -172,8 +172,15 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
       });
 
       function sortSidebarGroups(sidebarData) {
-        const pages = sidebarData.childNodes.sort((a, b) => a.priority - b.priority);
-        return { ...sidebarData, childNodes: pages };
+        const sortedGroupedPages = sidebarData.map(pageDir => {
+          if (pageDir.childNodes) {
+            const sortedPages = pageDir.childNodes.sort((a, b) => a.priority - b.priority);
+            return { ...pageDir, childNodes: sortedPages };
+          } else {
+            pageDir;
+          }
+        });
+        return sortedGroupedPages;
       }
 
       await Promise.all(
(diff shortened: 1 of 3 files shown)


packages/plugins/src/SidebarPlugin.ts
```diff
@@ -164,15 +152,29 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
         deep: sidebarRootLevel
       });
 
+      //Map into Sidebar Groups and sort children according to priority, priority = 1 has the highest priority
+      function sortSidebarGroups(sidebarData) {
+        const sortedGroupedPages = sidebarData.map(page => {
+          if (page.childNodes) {
+            const sortedPages = page.childNodes.sort((a, b) => a.priority - b.priority);
```
lukeac123, line 159: Priority = 1 now has the highest priority. Think this is most intuitive and it aligns with how the priority ranking for the plugins work

mark-tate, line 159: looks like we might be changing the way that priority works. Consider that this means all the existing pages will need changing, so this becomes a breaking change and something which need communication as well as codemods.

Instead we could create an additive change and deprecate the old
priority. JP used to talk about creating a ‘sidebar-> sortLabel’ instead of using a numeric value

The value here is that it would be less brittle and more flexible as you would not need to change the other numbers if new entries are added.

One to discuss with wider team, get someone’s else’s view and see what the consensus is.

ie. folks tend to add ascending integers for priority when all they want to do is set the priority of one label

ie.. Say you want to order 3 labels, instead of priority 1,2.3 you could just add one ‘sortLabel’ which ensures that the label appears in the right position

packages/site-components/src/Sidebar/styles.css.ts
```diff
@@ -4,6 +4,7 @@ import { sidebar } from '@jpmorganchase/mosaic-theme';
 export default {
   root: style([
     {
+      width: '400px',
```
DavieReid, line 7: Does this impact the responsive design in any way?  When I looked the sidebar goes full width on smaller devices so I guess not?

DavieReid (approved):

packages/site-components/src/Sidebar/styles.css.ts
mark-tate: yes , widths are driven via theme, so adding this might have some impact.

mark-tate, line 7: the left and right styles control tell the width

PR #230, opened 2023-02-23
Search results-list truncation

tomhazledine:
Searches with lots of results would create a list that overflowed the bottom of the page.
This change forces the results list to only ever show a maximum of 10 results at once.
Adds a "plus {N} more..." message to show how many results a search has found.

For this PR, `MAX_RESULTS` const is hardcoded to be 10, but the plan is to expose this into the options settings for the plugin (coming in a future PR)

### DIFF
--- a/packages/site-components/src/SearchInput/index.tsx
+++ b/packages/site-components/src/SearchInput/index.tsx
@@ -27,8 +27,8 @@ export function SearchInput() {
   useEffect(() => {
     const keys = ['title', 'content'];
     const results = performSearch(searchIndex, searchTerm, keys);
-    setListVisibility(true);
     setSearchResults(results);
+    setListVisibility(true);
   }, [searchTerm]);
 
   const handleInputFocus = () => {
--- a/packages/site-components/src/SearchInput/Results.tsx
+++ b/packages/site-components/src/SearchInput/Results.tsx
@@ -25,6 +25,8 @@ interface ResultListItemProps {
   onSelect: (result) => void;
 }
 
+const MAX_RESULTS = 10;
+
 function ResultListItem({ result, query, onSelect }: ResultListItemProps) {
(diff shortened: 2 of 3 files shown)


lilyvc:
Are the results ordered by relevance, so the hidden ones will be less useful?

packages/site-components/src/SearchInput/index.tsx
```diff
@@ -22,13 +23,22 @@ export function SearchInput() {
   const handleClear = useCallback(() => {
     setSearchTerm('');
     setSearchResults([]);
+    setSearchOverflow(0);
   }, [searchTerm, searchResults]);
 
   useEffect(() => {
+    const MAX_RESULTS = 10;
```
DavieReid, line 30: move this outside the component

packages/site-components/src/SearchInput/index.tsx
```diff
@@ -11,6 +11,7 @@ export function SearchInput() {
   const { searchIndex } = useSearchIndex();
   const [searchTerm, setSearchTerm] = useState('');
   const [searchResults, setSearchResults] = useState<SearchResults>([]);
+  const [searchOverflow, setSearchOverflow] = useState<number>(0);
```
DavieReid, line 14: I don't think this needs to be in state, derive the overflow from the `searchResults` state

tomhazledine:
> Are the results ordered by relevance, so the hidden ones will be less useful?

Yep - already ordered by relevancy. 👍

packages/site-components/src/SearchInput/index.tsx
tomhazledine: Done. Will be superseded by a prop soon anyway, so only a temporary const

lilyvc (approved):

PR #235, opened 2023-02-24
[WIP] testing SSR sizing for header icons

tomhazledine:


packages/theme/src/responsive/vars.css.ts
```diff
@@ -10,6 +10,7 @@ export const space = {
   x2: spaceScale(2),
   x3: spaceScale(3),
   x4: spaceScale(4),
+  x5: spaceScale(5),
```
mark-tate, line 13: FYI: Appreciate we are just experimenting.

I have asked about this a few times but was told the design folks don't want it as it would create odd spacing that creates centering issues.

Looks like we are using it for sizing icons which should use the `icon` semantic that already manages the icon sizing.

tomhazledine: Agree with those points. This was just a spike to see how easy it would be to clean up the first-load flash of SSR styles.

The real issue is that *no* Salt styles are being included in the SSR version. If we fix that, we won't need any "bodges" like this

mark-tate: MUI used to require us to add code to _document.js to load styles and avoid the flicker. Currently , Next loads styles from _app.js and we use that to load the Salt theme CSS, but that does not include the component css.

You can customise renderPage to inject the styles but we still have the problem of what we inject and where it comes from.

Might be worth conferring with @joshwooding as it's probably something we have not looked at yet with Salt.

Issue #155, opened 2023-01-15
SharedConfig must exist or a 404 is thrown


mark-tate:
*ACTUAL  RESULTS*
When creating a 1 page (or more)  file-system containing just an `index.mdx`, you have to specify

```
frameOverrides:
    header:
```

(or equivalent) to ensure a `sharedConfig` file exists.

Without this entry, then you get `404` errors.

*EXPECTED RESULTS*

A 404 error is not thrown.
If no `sharedConfig` is defined then it's not an error.

As part of fixing this error we should understand what the minimum frontmatter for a page is.
In theory it could be 

```
---
---
```

PR #239, opened 2023-02-26
Issue 155

mark-tate:
Addresses https://github.com/jpmorganchase/mosaic/issues/155


lilyvc (approved):

packages/site-middleware/src/withSearchIndex.ts
```diff
@@ -63,47 +30,53 @@ export const withSearchIndex: MosaicMiddleware<SearchIndexSlice> = async (
...
-      if (response.ok && response.status !== 204) {
-        searchIndexData = await response.json();
-      } else {
-        const show500 = response.status !== 404 && response.status !== 204;
-        const show404 = response.status === 404;
-        let errorMessage = `Could not find any search data defined for ${searchDataUrl}`;
-        if (show500) {
-          errorMessage = `An un-expected error occurred reading ${searchDataUrl}`;
+      const mosaicUrl = res.getHeader('X-Mosaic-Content-Url');
+      const response = await fetch(`${mosaicUrl}${urlPath}/search-data.json`, {
```
tomhazledine, line 59: `urlPath` doesn't apply for search data. Unlike sharedConfig, `search-data.json` is created once at the project root, so this line should be 
```
const response = await fetch(`${mosaicUrl}/search-data.json`, {
```

The also applies to the snapshot/S3 namespaces. As written, they will all 404

Search-data is global for the entire site, and my plan was to implement namespace support within that single search-data blob (i.e. `searchData = {namespaceOne: [/*content*/],namespaceTwo: [/*content*/]}`)

It might actually make more sense to handle the namespaced search-data in separate files as proposed here, but that would mean changing the searchIndex plugin too.

tomhazledine (commented):
Love the general gist of these changes, but they do break the search functionality (see note on `packages/site-middleware/src/withSearchIndex.ts`)

My recommendation would be to remove namespace support from search for this PR and merge the rest, then tackle search namespacing separately

packages/site-middleware/src/withSearchIndex.ts
```diff
@@ -63,47 +30,53 @@ export const withSearchIndex: MosaicMiddleware<SearchIndexSlice> = async (
...
-    let searchIndexData;
+    let searchIndex;
     if (isSnapshotFile) {
-      const env = getEnv(urlPath);
-      const { MOSAIC_SNAPSHOT_DIR: mosaicSnapshotDir } = env.data;
-      const filePath = path.join(process.cwd(), mosaicSnapshotDir, 'search-data.json');
-      const rawSearchIndexData = await loadLocalFile(filePath);
-      searchIndexData = JSON.parse(rawSearchIndexData);
+      const { snapshotDir } = getSnapshotFileConfig(urlPath);
+      const filePath = path.join(process.cwd(), snapshotDir, urlPath, 'search-data.json');
```
tomhazledine, line 37: Temp fix to revert namespace support

```suggestion
      const filePath = path.join(process.cwd(), snapshotDir, 'search-data.json');
```

packages/site-middleware/src/withSearchIndex.ts
tomhazledine: Pushed a fix and amended the tests.

lukeac123 (approved):

PR #240, opened 2023-02-28
ensure the square secondary button use consistent styling driven by theme

mark-tate:


packages/theme/src/button/button.css.ts
```diff
@@ -2,12 +2,20 @@ import { style } from '@vanilla-extract/css';
...
 
-import { responsiveConditions, responsiveSprinkles } from '../responsive';
+import {breakpoint, responsiveSprinkles} from '../responsive';
 import { vars } from '../vars.css';
 import { ctaColorSprinkles, primaryColorSprinkles, secondaryColorSprinkles } from '../color';
 
 export const buttonSizeProperties = defineProperties({
-  conditions: responsiveConditions,
+  conditions: {
+    mobile: { '@media': `screen and (min-width: ${breakpoint.mobile}px)`, selector: '.saltButton&' },
```
mark-tate, line 11: note: how you can add a selector to a condition to make it more specific

packages/theme/src/button/button.css.ts
```diff
@@ -102,23 +105,22 @@ export const button = recipe({
       square: style([
```
mark-tate, line 105: this style was unused before but it was intended to be used for the square buttons, like we see in app header and drawer etc.

packages/theme/src/button/vars.css.ts
```diff
@@ -7,10 +7,10 @@ const defaultButtonVars = {
     height: '44px'
   },
   web: {
-    height: '28px'
+    height: '44px'
```
mark-tate, line 10: temp change that makes all button have consistent size.
Without this, you would have app header buttons 28px x 44px or we could make them 28px x 28px but then they do not highlight the whole app header on hover.

Once we have this deployed we can do a design review and reconciliation.
It feels like the IA around breakpoints needs review and then we update the styles as needed.

Some of the current issues with the spec.
- At mobile/tablet, the drawer can be opened from two places (top-left+bottom-right buttons). 
- The top-left switches from an open to a close button, what happens with the bottom-right button
- The drawer at mobile/tablet is not full width but it should maximise the use of space
- it's possible to have two icons on the screen that look the same and both open drawers
- the bottom-right hand button is somewhat hidden and hard to see, you have to know it is there

At present, styles are consistently driven by theme, so it's a known starting point which we can iterate from.

lukeac123 (approved):

PR #241, opened 2023-03-02
Sidebar priority2

lukeac123:
- Add recursive sort function that orders the pages based on the sidebar priority
- Function moved to after the pages are grouped therefore the ./index page doesn't need to be pinned to the front (this code has been removed). This is because after the createGroupMap function the ./index pages are the parents and pages of the same length and routeDir are contained in the ChildNodes array

### DIFF
--- a/.changeset/happy-actors-cover.md
+++ b/.changeset/happy-actors-cover.md
@@ -1,0 +1,31 @@
+---
+'@jpmorganchase/mosaic-plugins': patch
+'@jpmorganchase/mosaic-cli': patch
+'@jpmorganchase/mosaic-components': patch
+'@jpmorganchase/mosaic-labs-components': patch
+'@jpmorganchase/mosaic-content-editor-plugin': patch
+'@jpmorganchase/mosaic-core': patch
+'@jpmorganchase/mosaic-create-site': patch
+'@jpmorganchase/mosaic-from-http-request': patch
+'@jpmorganchase/mosaic-layouts': patch
+'@jpmorganchase/mosaic-open-api-component': patch
+'@jpmorganchase/mosaic-schemas': patch
+'@jpmorganchase/mosaic-serialisers': patch
+'@jpmorganchase/mosaic-site': patch
+'@jpmorganchase/mosaic-site-components': patch
+'@jpmorganchase/mosaic-site-middleware': patch
+'@jpmorganchase/mosaic-site-preset-styles': patch
+'@jpmorganchase/mosaic-source-git-repo': patch
+'@jpmorganchase/mosaic-source-http': patch
+'@jpmorganchase/mosaic-source-local-folder': patch
+'@jpmorganchase/mosaic-standard-generator': patch
+'@jpmorganchase/mosaic-store': patch
+'@jpmorganchase/mosaic-theme': patch
+'@jpmorganchase/mosaic-types': patch
+'@jpmorganchase/mosaic-workflows': patch
+---
+
+1. Pip Salt version
+2. Issues 155, make 500 error more specific
+3. button and sidebar styles
+4. search opt-out
--- a/packages/plugins/src/SidebarPlugin.ts
+++ b/packages/plugins/src/SidebarPlugin.ts
@@ -154,7 +154,7 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
 
       //Map into Sidebar Groups and sort children according to priority
       function sortSidebarGroups(sidebarData) {
-        let sortedGroupedPages = sidebarData.map(page => {
+        const sortedGroupedPages = sidebarData.map(page => {
           if (page.childNodes.length > 1) {
             const sortedPages = page.childNodes.sort(
               (pageA, pageB) =>
--- a/packages/plugins/src/SidebarPlugin.ts
+++ b/packages/plugins/src/SidebarPlugin.ts
@@ -15,21 +15,6 @@ function createFileGlob(patterns, pageExtensions) {
   return `${patterns}{${pageExtensions.join(',')}}`;
 }
 
-function sortPagesByPriority(pageA, pageB, dirName) {
-  // Always pin /index to the front
-  const route = `${dirName}/index`;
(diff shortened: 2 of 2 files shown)


packages/plugins/src/SidebarPlugin.ts
```diff
@@ -40,8 +25,6 @@ function sortByPathLevel(pathA, pathB) {
   return pathBLevel - pathALevel;
 }
 
-const filterPages = page => !(page.sidebar && page.sidebar.exclude);
```
mark-tate, line 43: we need this code as some pages in Developer have exclude set, to not show them in the sidebar.
This is used when you want to hide pages.

You can create a page and have it as a loadable page, accessed from the full URL but not visible via the menus.

packages/plugins/src/SidebarPlugin.ts
```diff
@@ -172,14 +152,18 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
...
+          if (page.childNodes.length > 1) {
+            const sortedPages = page.childNodes.sort(
+              (pageA, pageB) =>
+                (pageB.priority ? pageB.priority : -1) - (pageA.priority ? pageA.priority : -1)
+            );
+            sortSidebarGroups(page.childNodes);
             return { ...page, childNodes: sortedPages };
           } else {
-            page;
+            return page;
```
mark-tate, line 166: good spot... this is why it's a good idea to add tests, we would of seen this issue sooner.
our test coverage is improving, so consider adding tests, as we go.

not to say we need to block this PR as it's missing tests but always nice to see a PR with test coverage.

packages/plugins/src/SidebarPlugin.ts
```diff
@@ -172,14 +152,18 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
...
 
-      //Map into Sidebar Groups and sort children according to priority, priority = 1 has the highest priority
+      //Map into Sidebar Groups and sort children according to priority
       function sortSidebarGroups(sidebarData) {
         const sortedGroupedPages = sidebarData.map(page => {
-          if (page.childNodes) {
-            const sortedPages = page.childNodes.sort((a, b) => b.priority - a.priority);
+          if (page.childNodes.length > 1) {
+            const sortedPages = page.childNodes.sort(
+              (pageA, pageB) =>
```
mark-tate, line 160: generally nicer to have a named sort function as this is then not an anonymous sort..
See how we named it sortByPathLevel or sortPagesByPriority

I don't have to read the code to know how the sort works.

packages/plugins/src/SidebarPlugin.ts
```diff
@@ -172,14 +152,18 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
...
+                (pageB.priority ? pageB.priority : -1) - (pageA.priority ? pageA.priority : -1)
```
mark-tate, line 161: the original sort would prioritise the index pages of the current directory/group
 this now just depends on priority
 
 Are the sorts equivalent, probably worth adding a description to this PR, so we can see what we changing.
 
 See @tomhazledine latest pr, which has a great overview of the search change

packages/plugins/src/SidebarPlugin.ts
```diff
@@ -172,29 +152,35 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
         deep: sidebarRootLevel
       });
 
-      //Map into Sidebar Groups and sort children according to priority, priority = 1 has the highest priority
-      function sortSidebarGroups(sidebarData) {
-        const sortedGroupedPages = sidebarData.map(page => {
-          if (page.childNodes) {
-            const sortedPages = page.childNodes.sort((a, b) => b.priority - a.priority);
-            return { ...page, childNodes: sortedPages };
+      const removeExcludedPages = page => !(page.sidebar && page.sidebar.exclude);
```
mark-tate, line 155: FYI for the future reference, we could simplify this again.

const removeExcludedPages = page => !(page?.sidebar?.exclude);

however, this code reduces down to a single expression and is nice and readable.
unlike sort which is multiline and you need to decode it.

So this is a good example of where, an inline anonymous function could work for filtering pages but a named function works better for sorting.

packages/plugins/src/SidebarPlugin.ts
```diff
@@ -172,29 +152,35 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
...
-        return sortedGroupedPages;
+        return pagesByPriority;
       }
 
       await Promise.all(
         rootUserJourneys.map(async dirName => {
           const sidebarFilePath = path.posix.join(String(dirName), options.filename);
-          const pages = await createPageList(dirName);
+          let pages = await createPageList(dirName);
+          pages = pages.filter(page => removeExcludedPages(page));
```
mark-tate, line 177: FYI... this could of been
          pages = pages.filter(removeExcludedPages);
          but actually a one line function does not have a huge amount of value, so we could inline instead, see other comment
          
          This is a good example of the two possible types of refactoring, in one instance, we could inline and the other we breakout complexity into a named function.

.changeset/happy-actors-cover.md
```diff
@@ -0,0 +1,31 @@
...
+'@jpmorganchase/mosaic-source-http': patch
+'@jpmorganchase/mosaic-source-local-folder': patch
+'@jpmorganchase/mosaic-standard-generator': patch
+'@jpmorganchase/mosaic-store': patch
+'@jpmorganchase/mosaic-theme': patch
+'@jpmorganchase/mosaic-types': patch
+'@jpmorganchase/mosaic-workflows': patch
+---
+
+1. Pip Salt version
```
mark-tate, line 28: Tip: imagine reading the changelog as someone who is not in the team and not familiar with the change

Changelogs are documentation and often viewed in retrospect, so handy to add as much info as possible.

mark-tate (approved):

PR #242, opened 2023-03-02
change `nextImageClassName` type to optional

Fercas123:
Currently an empty string would need to be passed to satisfy the type, making it optional so it can be used without an added class

### DIFF
--- a/packages/site-components/src/Image/index.tsx
+++ b/packages/site-components/src/Image/index.tsx
@@ -13,7 +13,7 @@ export type ImageProps = Omit<NextImageProps, 'src'> & {
   /**
    * class name to apply to the inner Next Image component
    */
-  nextImageClassName: string;
+  nextImageClassName?: string;
 };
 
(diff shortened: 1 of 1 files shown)


lukeac123:
looks good but need to change your git config so you're not using your jpm email address, git config --global user.email "MY_NAME@example.com"

origami-z:
> looks good but need to change your git config so you're not using your jpm email address, git config --global user.email "[MY_NAME@example.com](mailto:MY_NAME@example.com)"

The error is due to GH action not authenticated

mark-tate:
I believe it is because it’s a fork and not a clone and the fork has not setup the same tokens needed by the  hook

our teams are not setup correctly so emailed the org owners so this does not need to be a fork

highlights that we need to setup salt and mosaic so we can all
contribute easily and we can maintain the teams ourselves

mark-tate (approved):

PR #243, opened 2023-03-03
Search Improvements

tomhazledine:
Three things happening here:
1. Search Plugin refactor
2. Changing the way Pull Docs handles plugin definitions
3. Adding opt-out option to the search plugin

### Search Plugin refactor

This fixes a bug where the search index would only contain data for the most recently updated source. Now each source's data is parsed in `$beforeSend()` and added to the global config. Then `afterUpdate()` combines all of these into one file that gets saved to the filesystem.

Functionally, this makes the search box show results for every namespace, much like the V1 Developer site did before we introduced namespace filtering. Namespace filtering will come to this search too, but not in this PR (there are a lot of pieces to move around to make that happen - this PR just addresses the unwanted race condition between the sources).

### Changing the way Pull Docs handles plugin definitions

Pull Docs will now check the list of plugins for duplicates (based on the `modulePath` value) and will merge any duplicates that it finds. Precedence is given to the user-defined options. This allows users to change *only certain options* for a plugin that may be initialised elsewhere (for example, the search plugin is in the standard generator, but users will most likely want to tweak *some* of the relevancy options).

The Pull Docs plugin definition test has been amended to check for this scenario, too.

### Adding opt-out option to the search plugin

As mentioned above, the search plugin is part of the standard generator but certain users may want to turn off search (this is the case for the Salt site currently). The `disabled` search option allows them to do this.


packages/core/src/helpers/parsePluginModuleDefinitions.ts
```diff
@@ -0,0 +1,30 @@
...
+  const pluginsByPath = plugins.reduce((acc, plugin) => {
+    if (acc[plugin.modulePath]) {
+      acc[plugin.modulePath] = merge(acc[plugin.modulePath], plugin);
+      return acc;
+    }
+    acc[plugin.modulePath] = plugin;
+    return acc;
+  }, {});
+
+  const unbundled = Object.keys(pluginsByPath).map(path => pluginsByPath[path]);
```
mark-tate, line 28: NIT: unbundled nomenclature threw me a little

could use
``` Object.values(pluginsByPath); ```
perhaps ?

File: packages/plugins/src/SearchIndexPlugin.ts [hunk 2]
```diff
@@ -13,6 +13,7 @@ interface SearchIndexPluginOptions {
   maxLineLength?: number;
   maxLineCount?: number;
   keys?: string[];
+  disabled?: boolean;
```
mark-tate, line 16: should every plugin support disabled ?

is there a common interface for all plugins, if not something to think about, follow up on

packages/plugins/src/SearchIndexPlugin.ts
```diff
@@ -73,15 +74,21 @@ const optimizeContentForSearch = async ({
...
+   * all sources have been loaded).
+   */
+  async $beforeSend(
     mutableFilesystem,
-    { sharedFilesystem, serialiser, ignorePages },
-    { maxLineLength, maxLineCount, keys }
+    { config, serialiser, ignorePages },
+    { disabled, maxLineLength, maxLineCount, keys }
   ) {
+    if (disabled) return;
```
mark-tate, line 91: Should we filter the plugins and remove disabled plugins rather than have to add a guard to each plugin lifecycle ?

packages/core/src/helpers/parsePluginModuleDefinitions.ts
tomhazledine: Good spot. I'd got a bit tunnel visioned here, I think 😆 
Fixed.

packages/plugins/src/SearchIndexPlugin.ts
tomhazledine: Great suggestion. Have added this.

packages/plugins/src/SearchIndexPlugin.ts
tomhazledine: Yep, this is a good idea.
I've added `disabled` to the global `pluginModuleSchema` so it's now available to all plugins. Also lifted it up a level (from `pluginModuleSchema.options.disabled` to `pluginModuleSchema.disabled`) which gives us a cleaner schema and Zod validation.

packages/standard-generator/src/fs.config.js
```diff
@@ -36,7 +36,7 @@ module.exports = {
     },
     {
       modulePath: '@jpmorganchase/mosaic-plugins/SearchIndexPlugin',
-      options: { maxLineLength: 240, maxLineCount: 240, keys: ['description'] }
+      options: { maxLineLength: 240, maxLineCount: 240 }
```
mark-tate, line 39: does this need changing in site as well ?

packages/core/src/helpers/parsePluginModuleDefinitions.ts
```diff
@@ -0,0 +1,29 @@
...
+ * This is useful for when a plugin is defined in the standard
+ * generator but then also overridden by the user's config.
+ *
+ * @param plugins an array of plugin module definitions
+ * @returns an array of plugin module definitions with duplicates removed
+ */
+export default function parsePluginModuleDefinitions(
+  plugins: PluginModuleDefinition[]
+): PluginModuleDefinition[] {
+  const pluginsByPath = plugins.reduce((acc, plugin) => {
```
mark-tate, line 19: I guess this is where we add a filter for disabled plugins ?

or is there a better place ?

packages/core/src/PullDocs.ts
```diff
@@ -48,7 +53,8 @@ export default class PullDocs {
       this.#vfs,
       sharedFilesystem,
       // Refs and aliases should be applied after all other plugins, so we add them manually with a negative priority
-      plugins
+      parsePluginModuleDefinitions(plugins)
+        .filter((plugin: PluginModuleDefinition) => !plugin.disabled)
```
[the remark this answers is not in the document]
tomhazledine: Yep. Felt cleaner to add it here. Plus it emphasises that the internal plugins (`$TagPlugin`, `$CodeModPlugin`, `$AliasPlugin`, `$RefPlugin`) are added after the filtering has happened (we don't want to allow people to turn those off!)

packages/standard-generator/src/fs.config.js
tomhazledine: Nope, site will pick it up from here 👍

mark-tate: ah right

PR #248, opened 2023-03-09
Playwright

DavieReid:


packages/site/playwright.config.ts
```diff
@@ -0,0 +1,47 @@
...
+const config: PlaywrightTestConfig = {
+  // Timeout per test
+  timeout: 10 * 1000,
+  // Test directory
+  testDir: path.join(__dirname, 'e2e'),
+  // If a test fails, retry it additional 2 times
+  retries: 2,
+  // Artifacts folder where screenshots, videos, and traces are stored.
+  outputDir: 'test-results/',
+
```
mark-tate, line 20: what’s our thoughts on storage 

maybe for now we can just use without screenshots

mark-tate (approved):

packages/site/playwright.config.ts
mark-tate: could also use chromatic

DavieReid: This looks to be the most feasible option for image storage https://git-lfs.com/

PR #251, opened 2023-03-11
Fix next prev

mark-tate:


packages/standard-generator/src/fs.config.js
```diff
@@ -55,16 +55,6 @@ module.exports = {
         cacheDir: '.tmp/.pull-docs-last-page-plugin-cache'
       }
     },
-    {
-      modulePath: '@jpmorganchase/mosaic-plugins/NextPrevPagePlugin',
-      options: {
```
DavieReid, line 60: I don't think these options have been ported to the Sidebar plugin.  Would be nice to keep them in some form...also add an option for `rootLevel`

packages/plugins/src/SidebarPlugin.ts
```diff
@@ -43,6 +43,11 @@ interface SidebarPluginOptions {
 /**
  * Sorts the pages in a folder by priority and then exports a JSON file (name: `options.filename`) with the
  * sidebar tree from that directory downwards and adds sidebar data into frontmatter for each page.
+ *
+ * Additionally, add to frontmatter
```
DavieReid, line 47: Nice

DavieReid (approved):

PR #264, opened 2023-03-14
enable sidebar root(s) to be configured

mark-tate:
using `SidebarPlugin` options, add `rootDirGlob`

`rootDirGlob` is a glob pattern for matching directories which should be the root of the sidebar

`**` creates a root level sidebar containing all pages 
`*/*` creates a sidebar under each of the root directories


mark-tate:
need to test this on Windows before merge (will do that next)

DavieReid (approved):

PR #265, opened 2023-03-15
Search config

tomhazledine:
Fuse.js (the matching engine used by Mosaic's search function) accepts a lot of options that have an impact on the relevancy of search results. These will often need to be fine-tuned on a site-by-site basis, so this PR allows users to set their relevancy options when setting the general `SearchIndexPlugin` options.

* Adds `relevancy` object to SearchIndexPlugin options
* Generates a `search-config.json` file in the virtual filesystem
* Pipes the search config into the site's store via middleware
* Adds tests for SearchIndexPlugin (specifically around how `keys` and `content` are parsed)

---

*__Note__: This is another step towards removing the search index data from the site's first load. That will come in a future PR*


packages/plugins/src/SearchIndexPlugin.ts
```diff
@@ -9,18 +9,70 @@ type LeafNode = Node & {
...
+ * There are three scenarios for the "keys":
+ *
+ * 1. No keys are specified in the plugin options. In this case, we'll use the fallback keys.
+ * 2. The keys are specified as an array of strings. In this case, we'll add the required keys to the array.
+ * 3. The keys are specified as an array of objects. In this case, we can assume "weighting" is being applied
+ *    so we'll only return the keys that have been specified in the options.
+ *
+ * @param optionKeys array of keys specified in the plugin options
+ * @returns array of keys to be used in the search index
+ */
```
DavieReid, line 71: Maybe we could move these comments above `key` property in the `SearchIndexPluginOptions` interface so the comment shows up in the IDE

packages/plugins/src/SearchIndexPlugin.ts
```diff
@@ -86,7 +143,7 @@ const SearchIndexPlugin: PluginType<Page, SearchIndexPluginOptions> = {
   async $beforeSend(
```
DavieReid, line 143: Interested to know why `$beforeSend` was picked instead of `$afterSource`.  I am trying to get document why you would pick one over the other...

packages/plugins/src/__tests__/SearchIndexPlugin.test.ts
```diff
@@ -0,0 +1,73 @@
+import { optimizeContentForSearch, parseKeys } from '../SearchIndexPlugin';
+
+describe('GIVEN the SearchIndexPlugin', () => {
```
DavieReid, line 3: Nice

packages/store/src/useSearchIndex.ts
```diff
@@ -4,11 +4,14 @@ export type { SearchIndexSlice } from './types';
 
 export function useSearchIndex() {
```
DavieReid, line 5: is the name still appropriate now that it returns the index and the config?

packages/plugins/src/SearchIndexPlugin.ts
```diff
@@ -9,18 +9,70 @@ type LeafNode = Node & {
...
 };
 
+interface IndexItem {
+  title: string;
+  route: string;
+  content?: string[];
+  [key: string]: string | string[];
+}
+
+interface SearchRelevancy {
```
DavieReid, line 19: Can this type be imported from types/SearchIndex?  They look the same but have a different name...

export type SearchConfig = {
  includeScore?: boolean;
  includeMatches?: boolean;
  maxPatternLength?: number;
  ignoreLocation?: boolean;
  threshold?: number;
  keys?: string[] | { name: string; weight: number }[];
};

tomhazledine: Good point. Added `SearchConfig` to types package.

packages/store/src/useSearchIndex.ts
tomhazledine: Switched to `useSearch()` 👍

PR #266, opened 2023-03-15
Update e2e

tomhazledine:
* updates page titles used in e2e tests
* fixes one instance of multiple elements being caught by a selector
* regen snapshot


DavieReid (approved):
Thanks for fixing this...

Not caused by this PR but highlighted by it...we have 2 snapshots of the same content.  Let's pick 1 and go with that

PR #267, opened 2023-03-15
Fix getProjectNameAndRepoName not working on Windows

joshwooding:


packages/source-git-repo/src/Repo.ts
```diff
@@ -20,14 +20,14 @@ export type DiffResult = Array<{
 }>;
 
 function getProjectNameAndRepoName(repoUrl: string) {
-  const [, projectNameAndRepoName] = path
-    .normalize(repoUrl)
-    .match(/([^/]+\/[^/]+)\.git$/) as RegExpMatchArray;
-  const parts = projectNameAndRepoName.split('/');
+  const [, projectNameAndRepoName] = new URL(repoUrl).pathname.match(
```
joshwooding, line 23: repoUrl is a url anyway so it should be treated like one.

DavieReid (approved):

PR #271, opened 2023-03-17
Theme variable names

DavieReid:


.changeset/mighty-tips-attend.md
```diff
@@ -0,0 +1,55 @@
+---
+'@jpmorganchase/mosaic-theme': patch
```
libertymayc, line 2: lgtm, but I think this may constitute more than a patch?

mark-tate: we are on beta numbering at the mo, so we have just been using patch for everything

mark-tate (approved):

.changeset/mighty-tips-attend.md
DavieReid: Agreed...updated the changset

PR #277, opened 2023-03-20
Chore changes

mark-tate:


package.json
```diff
@@ -9,7 +9,7 @@
     "packages/*"
   ],
   "engines": {
-    "node": "^14.18.0 || >=16.10.0 || >=18.0.0"
```
DavieReid, line 12: We can't do this at the moment

DavieReid: Ok, do you not need to make this change in all packages...

DavieReid (approved):

Issue #275, opened 2023-03-20
Vercel Deployment - No Content


DavieReid:
## Issue
When deploying a mosaic site and snapshot to vercel, the snapshot directory is not included in the code deployed to serverless functions.

This results in the site showing the 404 page for all content and the following is displayed in the vercel logs

```
ENOENT: no such file or directory, realpath '/var/task/<site-name>/public/snapshots/latest/<page-name>.mdx'
```

## Cause
After investigation, it appears that unless a file is somehow included in the source code, as described [here](https://vercel.com/guides/how-can-i-use-files-in-serverless-functions) then vercel will not include those files in the build output.

## Why does the mosaic site itself work?
The site middleware package includes code that resolves the files when using snapshot mode.

PR #278, opened 2023-03-21
Vercel snapshot deployments

DavieReid:


DavieReid:
Fixes - https://github.com/jpmorganchase/mosaic/issues/275

packages/cli/src/index.ts
```diff
@@ -20,30 +21,28 @@ program.parse();
...
+  throw new Error(`[Mosaic] no config file provided`);
+}
+
+if (program.args[0] === 'build') {
+  build(config.default, path.resolve(process.cwd(), options.out), options);
+}
+if (program.args[0] === 'serve') {
+  serve(config.default, options.port, options.scope && options.scope.split(','));
+}
+if (program.args[0] === 'upload') {
```
mark-tate, line 43: @DavieReid upload and deploy are the same thing, albeit on different platforms ?

mark-tate: Small tidy up required?
updateTraceFile feels vercel specific..
the current `upload-snapshot-s3` is equally s3 specific.
Do we just rename these files and call a 'deploy' method or similar?

packages/cli/src/vercel-snapshot.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Vercel uses output file tracing (https://nextjs.org/docs/advanced-features/output-file-tracing) to determine what files need to be deployed.
+ * Due to the dynamic nature of mosaic sites, files that are required, the snapshots, are not traced correctly.
+ *
+ * This script adds an entry into the [...route].js.nft.json file for each file in the provided snapshot ensuring that the snapshot files are deployed.
+ *
+ */
+export async function updateTraceFile(config, options) {
+  const { mode, platform } = config.deployment;
+
+  if (mode === 'snapshot-file' && platform === 'vercel') {
```
mark-tate, line 17: should we move this check upto index.ts or add code to prevent a silent fail

packages/cli/src/vercel-snapshot.ts
```diff
@@ -0,0 +1,53 @@
...
+    const buildDir = path.posix.join(projectBase, '.next');
+    const snapshotDir = options.out;
+    const snapshotName = options.name ?? new Date().toISOString();
+
+    if (!fsExtra.existsSync(buildDir)) {
+      console.warn(
+        '[Mosaic] unable to update nextjs trace file.  Build the site before generating the snapshot'
+      );
+    }
+    const nftFilePath = path.posix.join(buildDir, 'server', 'pages', '[...route].js.nft.json');
```
mark-tate, line 32: how did you work this out.. very impressive

packages/schemas/src/MosaicConfigSchema.ts
```diff
@@ -1,10 +1,12 @@
 import { z } from 'zod';
 import { fileExtensionSchema } from './fileExtensionSchema.js';
+import { modeSchema } from './modeSchema.js';
 import { pluginModuleSchema } from './PluginModuleSchema.js';
 import { serialiserModuleSchema } from './SerialiserModuleSchema.js';
 import { sourceModuleSchema } from './SourceModuleSchema.js';
 
 export const mosaicConfigSchema = z.object({
+  deployment: z.object({ mode: modeSchema, platform: z.string() }).optional(),
```
mark-tate, line 9: I guess we could be opinionated on platform values , rather than just string ?

packages/schemas/src/modeSchema.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod';
+
+export const modeSchema = z.union([
+  z.literal('mode'),
```
mark-tate, line 4: should mode be active ?

mark-tate (approved):
a few comments added

packages/cli/src/index.ts
DavieReid: Yeah we should revisit this...

packages/schemas/src/modeSchema.ts
DavieReid: yeah not sure what happened here...will update

packages/cli/src/vercel-snapshot.ts
DavieReid: https://nextjs.org/docs/advanced-features/output-file-tracing

> During next build, Next.js will use [@vercel/nft](https://github.com/vercel/nft) to statically analyze import, require, and fs usage to determine all files that a page might load.

My theory was, the mosaic architecture can't be statically analysed...you need to actually run the app, navigate to a route before the snapshot mdx page is read.  So therefor there will be no snapshot mdx files listed in the trace.

packages/schemas/src/MosaicConfigSchema.ts
DavieReid: fixed

PR #289, opened 2023-03-27
chore: run playwright tests on macos

DavieReid:


tomhazledine (approved):
Assuming that this works, as the tests pass 😏 
Screenshot 2023-03-27 at 11 52 10
Screenshot 2023-03-27 at 11 52 16

PR #294, opened 2023-03-29
Mosaic Site sidebar root

lilyvc:
- mosaic config for the site has been changed so that the sidebarRootGlob is \*/\*
- sidebar plugin has been updated to stop it throwing errors if childNodes is undefined
- snapshot has been regenerated to include sidebar changes


packages/plugins/src/SidebarPlugin.ts
```diff
@@ -218,18 +223,20 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
...
-                (pageB.priority ? pageB.priority : -1) - (pageA.priority ? pageA.priority : -1)
-            );
-            sortPagesByPriority(page.childNodes);
-            return { ...page, childNodes: pagesByPriority };
-          } else {
-            return page;
-          }
-        });
+        const pagesByPriority = sidebarData
+          .filter(page => !!page) // filter out null or undefined values
```
mark-tate, line 227: this feels like we might be filtering out null or undefined values but should they be there, to begin with?
We are potentially masking an error and not fixing it?

here is a thought.

If the pageExtensions is not set then would it glob every file in the directory, rather than just .mdx.
We have set the default value for pageExtensions to [], so it does not error but would glob every file and not just mdx.

Then if you have a file that does not contain frontmatter (or fullPath) it would be undefined.

So we could check what the file glob returns from this file and ensure every file has a fullPath and is a mdx file.

I know JP was keen to not hardcode mdx as a default type as they could be markdown.
Potentially something like  is needed in our standard generator configs`.{md,mdx}`

packages/site/mosaic.config.js
```diff
@@ -1,9 +1,14 @@
...
 const mosaicConfig = require('@jpmorganchase/mosaic-standard-generator/dist/fs.config.js');
 
-module.exports = deepmerge(mosaicConfig, {
-  deployment: { mode: 'snapshot-file', platform: 'vercel' },
+const siteConfig = {
+  ...mosaicConfig,
   plugins: [
+    ...mosaicConfig.plugins,
+    {
+      modulePath: '@jpmorganchase/mosaic-plugins/SidebarPlugin',
```
mark-tate, line 9: this should be set in the standard generator config rather than overriden.
The defaults should match what Salt needs, I would propose we design the Mosaic site with the same sidebar IA (or as close as possible).

packages/plugins/src/SidebarPlugin.ts
mark-tate: This is all theoritical ... but if true,  explains why sometimes it works and sometimes it does not, because it depends what files are in the directory being globbed.

lilyvc: > This is all theoritical ... but if true, explains why sometimes it works and sometimes it does not, because it depends what files are in the directory being globbed.

Extensions is set in the source, do you think it is not being applied?

PR #298, opened 2023-03-30
fix: prevent content section shrinking on page change

DavieReid:


packages/layouts/src/LayoutColumns/styles.css.ts
```diff
@@ -9,16 +9,16 @@ import {
 import { defineProperties, createSprinkles } from '@vanilla-extract/sprinkles';
 
 const gridTemplates = {
-  wide: `minmax(${vars.space.none}, 500px) minmax(auto, ${config.main.wideWidth}px) minmax(${vars.space.none}, 500px)`,
-  small: '1fr'
+  columns: `minmax(250px, 1fr) minmax(min-content, ${config.main.wideWidth}px) minmax(150px, 1fr)`,
```
lilyvc, line 12: is it possible to change this to at least match Salt? https://www.saltdesignsystem.com/getting-started/developing

columns: `minmax(300px, 1fr) minmax(min-content, ${config.main.wideWidth}px) minmax(282px, 1fr)`

DavieReid: yes, will do that now

lilyvc (approved):

PR #305, opened 2023-04-03
Update sidebar root

lilyvc:
This PR changes the sidebar from showing all pages, to showing all the pages one level down from the root (\*/\*).

Index pages have been added to the test and layout folders, as the sidebar plugin currently needs them to generate the sidebar correctly.


DavieReid:
> @DavieReid I'm not sure how to update the screenshots for the test, but once we do + this change is in then the sidebar won't change every time the content is updated.

I usually:

1. Spin up the site on localhost:3000
2. cd in packages/site
3. run `npx playwright test --update-snpashots`

DavieReid (approved):

DavieReid:
> > > @DavieReid I'm not sure how to update the screenshots for the test, but once we do + this change is in then the sidebar won't change every time the content is updated.
> > 
> > 
> > I usually:
> > 
> > 1. spin up the site on localhost:3000
> > 2. open another terminal window/tab
> > 3. cd into packages/site
> > 4. run `npx playwright test --update-snpashots`
> 
> thanks :) I did this - but it's still failing. In the 'actual' result the 'log in to comment' button is in the middle of the page..have you had this problem before?

No never had that...when I checked the snapshots the headings move very so slightly...might need to relax the diff settings

PR #308, opened 2023-04-04
Flatten sidebar

lilyvc:
PR with 2 changes
1. Make the first page of the sidebar the same level as its children.

|before|after|
|------|-----|
|Screenshot 2023-04-04 at 17 45 15|Screenshot 2023-04-04 at 17 46 31|

2.  Make the sidebar collapsed by default.


packages/plugins/src/SidebarPlugin.ts
```diff
@@ -235,6 +235,14 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
         return pagesByPriority;
       }
 
+      function moveRootPageDown(pagesByPriority) {
```
mark-tate, line 238: We are modifying the data to fix a styling issue, which is not an ideal situation.
Someone would see it and might think it's a bug as no docs here, as to why we are doing it.

I wonder, if it's possible to do this without adding a root page as a child?

If we kept this code as it was. 
Do we just need to remove the margin from the first element in the vertical navigation?
We can then just modify the Vertical Navigation and keep the integrity of the data, so other sidebars don't have to treat element 0 as a special case.

As this is currently pro-sidebar.. suspect this is possible ?

lilyvc: > We are modifying the data to fix a styling issue, which is not an ideal situation. Someone would see it and might think it's a bug as no docs here, as to why we are doing it.
> 
> I wonder, if it's possible to do this without adding a root page as a child?
> 
> If we kept this code as it was. Do we just need to remove the margin from the first element in the vertical navigation? We can then just modify the Vertical Navigation and keep the integrity of the data, so other sidebars don't have to treat element 0 as a special case.
> 
> As this is currently pro-sidebar.. suspect this is possible ?

I don't agree that it is a styling issue. It is not just that the first element has a different margin, it is because it shouldn't be the parent of the other pages.

In a folder like:

author/
|─ index.mdx/
|─ layouts.mdx
|─ modes-of-operation/
     |─ active-mode.mdx
     |─ index.mdx

we want the sidebar tree to look like this:

author
layouts
modes of operation
     active mode

currently it is returning:

author
  layouts
  modes of operation
       active mode

I think the sidebar plugin should return the correct json, rather than configuring the vertical navigation to treat the first item differently.

mark-tate: interesting..what if we had a global sidebar created by using the glob '*' for root directories ?
That is what the F product wants I think ?

lilyvc: > interesting..what if we had a global sidebar created by using the glob '*' for root directories ? That is what the F product wants I think ?

it would previously have been (using mosaic site as an example)

mosaic
─ quick start guides
─ ─ publish to aws
─ configure
─ ─ sources
─ ─ ─ local source
─ ─ ─ github repo
......

with these changes, every other page except mosaic is one level less indented so it would be like:

mosaic
quick start guides
─ publish to aws
configure
─ sources
─ ─ local source
─ ─ github repo
......

DavieReid:
I think this is OK for now but wanted to check that we can still have the full expanded sidebar?  Is that just a matter of changing the glob?

lilyvc:
> I think this is OK for now but wanted to check that we can still have the full expanded sidebar?  Is that just a matter of changing the glob?

If you don't configure the glob in the mosaic config file then it will default to the full sidebar of the whole site

It won't be visually expanded by default though (meaning the sub menus are not visible until you expand them) but I think that makes sense because you can't really read a sidebar with 100 items, makes sense to just expand the sections you're interested in.

lukeac123:
Yeh i agree with the new structure and it not being expanded by default. Also a thought, do we want the parent dir e.g "Mosaic" or "Developer Home" to be in the sidebar ?
Surely the user knows their int he website and can use the logo in the app header to navigate back to the home page

lukeac123 (approved):

PR #311, opened 2023-04-11
310 sources should work without prefixdir

DavieReid:


packages/plugins/src/LazyPagePlugin.ts
```diff
@@ -66,6 +66,8 @@ const LazyPagePlugin: PluginType<LazyPagePluginPage, LazyPagePluginOptions> = {
       cwd: '/',
       onlyFiles: true
     });
+
+    console.log('lazy beforeSend');
```
DavieReid, line 70: will remove

DavieReid: DONE

packages/site/mosaic.config.js
```diff
@@ -38,12 +38,12 @@ module.exports = deepmerge(siteConfig, {
      * Access from your browser as http://localhost:3000/mosaic
      */
     {
-      disabled: process.env.NODE_ENV !== 'development',
+      // disabled: process.env.NODE_ENV !== 'development',
```
mark-tate, line 41: commented out code

packages/site/mosaic.config.js
```diff
@@ -7,7 +7,7 @@ const siteConfig = {
     ...mosaicConfig.plugins,
     {
       modulePath: '@jpmorganchase/mosaic-plugins/SidebarPlugin',
-      options: { rootDirGlob: '*/*' }
+      options: { rootDirGlob: '*' }
```
mark-tate, line 10: this flips us back to a single sidebar for everything..
I think we were going to converge with Salt.. is this a change we still want ?

packages/plugins/src/BreadcrumbsPlugin.ts
```diff
@@ -24,7 +24,7 @@ const BreadcrumbsPlugin: PluginType<BreadcrumbsPluginPage, BreadcrumbsPluginOpti
         let currentPage = page;
         let parentDir = path.posix.normalize(path.posix.dirname(currentPage.fullPath));
 
-        while (currentPage !== undefined) {
+        while (currentPage !== undefined && currentPage.fullPath !== '/index.mdx') {
```
mark-tate, line 27: I remember JP be vocal about not coupling us to specific file types ?
Some of our configs have things like 'extensions' in them, so hardcoding a filetype in source, feels very specific ?
Maybe our position has changed on this but what's the rule going forward and how does this affect config values ?

packages/site/mosaic.config.js
DavieReid: There is no sidebar without this...

packages/plugins/src/BreadcrumbsPlugin.ts
DavieReid: Will rework

DavieReid: Fixed

packages/site/mosaic.config.js
DavieReid: removed this code...it was only for testing

packages/site/next.config.js
```diff
@@ -50,17 +50,7 @@ module.exports = {
...
-      },
-      {
-        source: '/mosaic',
-        destination: '/mosaic/index',
-        permanent: true
-      },
-      {
-        source: '/local',
-        destination: '/local/index',
+        destination: '/index',
```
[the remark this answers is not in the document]
DavieReid: removed this code...it was only for testing

packages/site/mosaic.config.js
DavieReid: removed this code...it was only for testing

lilyvc (approved):

PR #312, opened 2023-04-11
test: add e2e tests for aliases

DavieReid:


tomhazledine:
The alias tests passed locally for me (with the caveat that I'm still commenting out `disabled: process.env.NODE_ENV !== 'development'` from my config)

PR #318, opened 2023-04-13
Missing index page warning

lukeac123:
Problem Statement:
- When a user doesn't include an index page in a DIR the page is returning undefined
- This causes issues with the sidebar and next / prev components

Solution 1: (As seen below)
- We notify users that there are index pages missing in some DIRs
- We still let the site to build but users will see issue when they navigate to a page with a sidebar (We could throw and error and prevent the site from building)

Solution 2:
- Insert code to assign a one of the pages as an Index page
- This changes the 'full path' of the page and could therefore cause issues if that page is linked to in other pages

Solution 3:
- We enable users to have any page names they want 
- This will require the group map in sidebarPlugin and potentially other code changes where index page is used e.g default routing to add /index when the route page dir is hit

### DIFF
--- a/packages/plugins/src/SidebarPlugin.ts
+++ b/packages/plugins/src/SidebarPlugin.ts
@@ -204,6 +204,11 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
             }
             if (currPage !== undefined) {
               createNavigationRefs(currPage, prevPage, nextPage);
+            } else if (currPage === undefined) {
+              console.warn(
+                '\x1b[31m',
+                `*** PAGE RETURNING UNDEFINED -  PLEASE CHECK ALL YOUR DIR'S CONTAIN INDEX PAGES. THIS MAY CAUSE ISSUES WITH SIDEBAR AND NAVIGATION ***`
+              );
             }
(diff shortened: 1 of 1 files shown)


packages/plugins/src/SidebarPlugin.ts
```diff
@@ -204,6 +204,11 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
             }
             if (currPage !== undefined) {
               createNavigationRefs(currPage, prevPage, nextPage);
+            } else if (currPage === undefined) {
+              console.warn(
+                '\x1b[31m',
+                `*** PAGE RETURNING UNDEFINED -  PLEASE CHECK ALL YOUR DIR'S CONTAIN INDEX PAGES. THIS MAY CAUSE ISSUES WITH SIDEBAR AND NAVIGATION ***`
+              );
```
lilyvc, line 211: Might be good to make this more consistent with the style of other warnings like:

console.warn('Page returning undefined. Please ensure that all directories contain index pages. This may cause issues with sidebar and navigation.');

lilyvc: yeah probably best to throw an error so it will stop and then the user can read what the problem was

PR #319, opened 2023-04-13
warning for dir with missing index pages

lukeac123:
Problem Statement:

When a user doesn't include an index page in a DIR the page is returning undefined
This causes issues with the sidebar and next / prev components
Solution 1: (As seen below)

We notify users that there are index pages missing in some DIRs
We still let the site to build but users will see issue when they navigate to a page with a sidebar (We could throw and error and prevent the site from building)
Solution 2:

Insert code to assign a one of the pages as an Index page
This changes the 'full path' of the page and could therefore cause issues if that page is linked to in other pages
Solution 3:

We enable users to have any page names they want
This will require the group map in sidebarPlugin and potentially other code changes where index page is used e.g default routing to add /index when the route page dir is hit

### DIFF
--- a/packages/plugins/src/SidebarPlugin.ts
+++ b/packages/plugins/src/SidebarPlugin.ts
@@ -205,6 +205,12 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
             if (currPage !== undefined) {
               createNavigationRefs(currPage, prevPage, nextPage);
             }
+            if (currPage === undefined) {
+              console.warn(
+                '\x1b[31m',
+                `*** PAGE RETURNING UNDEFINED -  PLEASE CHECK ALL YOUR DIR'S CONTAIN INDEX PAGES. THIS MAY CAUSE ISSUES WITH SIDEBAR AND NAVIGATION ***`
+              );
+            }
             recursiveAddNavigation(page.childNodes);
           });
         }


DavieReid:
I don't see how this helps...

Fix the plugins so that users can have any page name they want.

packages/plugins/src/SidebarPlugin.ts
```diff
@@ -205,6 +205,12 @@ const SidebarPlugin: PluginType<SidebarPluginPage, SidebarPluginOptions, Sidebar
             if (currPage !== undefined) {
               createNavigationRefs(currPage, prevPage, nextPage);
             }
+            if (currPage === undefined) {
+              console.warn(
+                '\x1b[31m',
+                `*** PAGE RETURNING UNDEFINED -  PLEASE CHECK ALL YOUR DIR'S CONTAIN INDEX PAGES. THIS MAY CAUSE ISSUES WITH SIDEBAR AND NAVIGATION ***`
```
lukeac123, line 211: console.warn - coloured in red and with *** to make it stand out. No index pages will cause pages with sidebar to fail so wanted to be very visible in the log

mark-tate: capitals is rather shouty..

not having an index is a valid, if we want hidden page support

who discovers the error, the user or the site maintainer ?
If the build fails, the site maintainer does
If the page fails the user does

Worth a discussion on the problem statement to understand what solution is best

formatted logs don’t work so well in splunk 

Might be worth considering this in a wider context, how should we handle page errors in general ?

Next has a red box and overlay to report page errors. Typically markdown solutions never fail and always try to display content.

DavieReid:
Ideal scenario is to avoid the need to create an index file in every directory.  Would this work:

Filesystem:

Folder A
   Doc 1
   Doc 2
   Doc 3


1. Order the docs within Folder A by sidebar priority with the highest priority being the "first" page.
2. If there is no sidebar priority then pages are sorted alphabetically
3. Check each of the pages for new frontmatter `sidebarGroupLabel` 
4. If found then that is used as the sidebar group name
5. Clicking a group label navigates to the first page
6. If no sidebar group label is found then use the name of the folder (?)

So we could have:

Folder A
   Doc 1 - priority 3
   Doc 2 - priority 2, groupLabel = MyDocs
   Doc 3 - priority 1

Produces the following sidebar
----------
MyDocs - clicking me does to Doc 1
   Doc 1
   Doc 2
   Doc 3

And without sidebar group label: 

Folder A - clicking me goes to Doc 1
   Doc 1
   Doc 2
   Doc 3

mark-tate:
is `sidebarGroupLabel`  the same things as saying.. "This is the index page for the group".
We already have `sidebarLabel` so having another property for the same thing seems duplication, do we just need something like `isIndex: true` ?

you can only have one 'isIndex' but it uses the first one, it finds in sidebar order.

OR 

Do we even need any metadata, if we have a plugin, which creates any missing index.mdx pages and sets them to a metadata driven redirect.

it always links to the first page, based on sidebar priority

PR #321, opened 2023-04-17
Add authoring docs

lilyvc:


docs/author/page-templates/docs/api-reference.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: API reference
+description: Replace this with a page description
```
mark-tate, line 3: should we remove description if we are not using it ?

DavieReid:
The Page templates section needs some work.  It throws 500 errors for nearly all of the pages and I don;t think the IA is right there.

For UI components, remove anything that does not come as part of standard markdown .e.g. Lists and Links.  We don't want people to use these when writing the docs, they should use markdown.

The Component example also blends in with the component purpose/description.  We may need a designer to look at how the example can be made more prominent. 

Maybe we need to start looking at something like sandpack for these examples.  The would be much more helpful if you can see the mdx for a callout, list, tile etc

DavieReid:
Reading through the page template section, i think this is internal stuff and shouldn't be here

lilyvc:
closing this for now - will do it in smaller iterations, adding the component docs first.

PR #325, opened 2023-04-18
Prevent duplication of search results content

tomhazledine:
When search results are displayed, they include the page route, the title of the page, and a one-line snippet of content.

When Fuse.js finds a match, it provides the line of content along with highlighting information (which characters should be bold). This is great when the "best match" on the page is part of the regular content, but leads to messy and confusing results when the best match is in either the title or route of the page: i.e. we get duplicated content being displayed, as well as *lower* information density (because we're losing the extra context provided by the one-line snippet.

This change makes search results snippet display the first line of the content when the "best match" is in the route or title (so a result will *always* include the route, title, and a useful text snippet).

Screenshot 2023-04-18 at 13 02 43

---

This PR also includes:
* types and tests for search utils.
* Reduced number of results shown on screen at any one time (`10` was overflowing the page on smaller screens. `6` feels more sensible)

### DIFF
--- a/packages/site-components/src/SearchInput/Results.tsx
+++ b/packages/site-components/src/SearchInput/Results.tsx
@@ -5,12 +5,7 @@ import { Caption6, P4, P6 } from '@jpmorganchase/mosaic-components';
 import { Highlighter } from '@salt-ds/lab';
 
 import styles from './styles.css';
-
-interface SearchResult {
-  title: string;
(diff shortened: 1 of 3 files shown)


mark-tate (approved):

mark-tate (approved):
could add a changeset to this, so that we incrementally create the release notes

PR #327, opened 2023-04-20
TOC Indentation

lilyvc:
This PR contains 5 commits corresponding to changes to the TOC.

1. Styling has been added so that TOC items are indented depending on the level of the heading.

2. The TOC plugin's minRank and maxRank have been updated to exclude H1s. This change was made because when the indentation was added, it became apparent that the H1 serves as the title of the page rather than part of its contents. Including it made the indentation look unusual when everything was nested under it.

3. Fix TOC item highlighting by including all headings in the page when finding the current heading. Prior to this fix, the highlighting was broken and did not correspond to the correct heading in the TOC. This was due to headings being removed from the TOC, which revealed the issue.

4. Update the snapshot. 

5. Change setupHeadingState test to reflect the change made in point 3 above.

### DIFF
--- a/packages/site-components/src/TableOfContents/TableOfContents.tsx
+++ b/packages/site-components/src/TableOfContents/TableOfContents.tsx
@@ -21,7 +21,7 @@ export const TableOfContents: React.FC<TableOfContentsProps> = ({ items }) => {
     throw new Error('No `items` specified for Table of Contents.');
   }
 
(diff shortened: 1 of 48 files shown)


mark-tate (approved):
could add a changeset so you write the release notes for this change, as part of this PR.

leave for the next pr though if you prefer, see how Salt Prs are created, they do that for each merge to main

PR #330, opened 2023-04-21
add siteUrl option to sitemap plugin

lilyvc:


packages/standard-generator/src/fs.config.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
   plugins: [
     {
       modulePath: '@jpmorganchase/mosaic-plugins/SiteMapPlugin',
-      options: {}
+      options: { siteUrl: process.env.SITE_URL || 'http://localhost:3000' }
```
mark-tate, line 35: does this need a doc change or read
e change ?

DavieReid: We should generate a new snapshot for the Mosaic site right?  Our sitemap will be not be correct atm.

lilyvc: yep I'll make sure this is explained in the docs and regen the sitemap

Although, I was having trouble with env vars not getting picked up. Where would we expect the Salt site for example to set the SITE_URL? Are they all stored in the env files or are we setting some through Vercel?

DavieReid: Yeah this is becoming more of a problem, notably with disabling of the sources.  We have .env files but I think they only apply to the nextjs site. 

We may now need a more comprehensive solution that uses dotenv so that both the site and mosaic fs can read env variables...might be one for another PR tho...

lilyvc: Okay cool, we are on the same page. I'll add dotenv in another PR as a prerequisite for this one.

PR #332, opened 2023-04-24
Add dotenv

lilyvc:
This PR adds the dotenv packages and uses it in mosaic config so that mosaic can access environment variables.

It defaults to reading '.env' but if you run the application with NODE_ENV=prod then it would read '.env.prod'.

I'm not really sure about this

-firstly because it seems strange to share the env files between the next.js application and the mosaic instance
-secondly because it seems strange to have to set NODE_ENV=local for the local development environment

Is there a better approach that I am missing?

### DIFF
--- a/packages/site/mosaic.config.js
+++ b/packages/site/mosaic.config.js
@@ -1,5 +1,8 @@
 const deepmerge = require('deepmerge');
 const mosaicConfig = require('@jpmorganchase/mosaic-standard-generator/dist/fs.config.js');
+const dotenv = require('dotenv');
+const envPath = process.env.NODE_ENV ? `.env.${process.env.NODE_ENV}` : '.env';
+dotenv.config({ path: envPath });
 
 const siteConfig = {
   ...mosaicConfig,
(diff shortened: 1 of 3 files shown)


tomhazledine:
Have used `dotenv` in the past. Kinda think of it as the "standard" for this king of thing, so no complains from me 👍

DavieReid:
> This PR adds the dotenv packages and uses it in mosaic config so that mosaic can access environment variables.
> 
> It defaults to reading '.env' but if you run the application with NODE_ENV=prod then it would read '.env.prod'.
> 
> I'm not really sure about this
> 
> firstly because it seems strange to share the env files between the next.js application and the mosaic instance secondly because it seems strange to have to set NODE_ENV=local for the local development environment Is there a better approach that I am missing?

Why do you need to use local for local development?  Because that's what we have currently in the site? `env.local`. Could create an env.development 1 or something like that.

I don't think we are sharing env files between the site and the fs.  We still have the config for the fs in the yarn command.  Things like port and path to the config file.  You could create env files in the cli package for each env.

BUT...

I still think there is the same issue: if the site runs in development mode...I don't think the fs knows that...  ( I haven't run it)

You may need something like https://github.com/entropitor/dotenv-cli

So that when you do `yarn serve` everything that runs knows what env it's running in.

DavieReid:
Also, seen people validate env files with a zod schema --> https://dev.to/jfranciscosousa/validating-environment-variables-with-zod-14am

PR #337, opened 2023-04-25
Fix Windows support

joshwooding:


packages/source-local-folder/src/index.ts
```diff
@@ -58,14 +58,14 @@ const LocalFolderSource: Source<LocalFolderSourceOptions> = {
           // eslint-disable-next-line @typescript-eslint/return-await
           await Promise.all(
             filepaths.map(async (filepath: string) => {
-              const fullPath = path.join(options.rootDir, filepath);
+              const fullPath = path.posix.join(options.rootDir, filepath);
```
joshwooding, line 61: Not 100% sure these changes are needed...

tomhazledine (approved):
Have confirmed everything still builds and runs on my MBP and on the PR preview (fresh `yarn` followed by `yarn build` then checked `yarn serve` and `yarn gen:snapshot` - all ran as-expected)

Haven't checked on Windows to confirm (but it can't be any *more* broken there 😆 )

Happy to tick as the changes make sense.

PR #343, opened 2023-04-28
Fix `document` undefined error

JoanaMMoreira:


lilyvc (approved):

DavieReid:
There is a few places where the TOC component uses `document`.  Maybe we just make this a client-side only component...

DavieReid:
Missing a changset

JoanaMMoreira:
> There is a few places where the TOC component uses `document`. Maybe we just make this a client-side only component...

I've tried that initially but it throws the same error `document is not defined` :(

tomhazledine (approved):

PR #346, opened 2023-05-02
fix: hide images when SSR class is applied

DavieReid:
style-inject is used by salt to insert styles.  This doesn't support SSR --> https://github.com/egoist/style-inject/blob/master/src/index.js#L2

So the styling that prevents images and icons from being huge is not applied.

This change adds a class before hydration that hides `img` and `svg` elements.

### WARNING:

This change dropped the Lighthouse perf metric from 98 down to 86.  I don't know if this is because it was being checked on a vercel preview branch rather than the main prod branch or if the change is the issue.


lilyvc (approved):
looks better :)

PR #355, opened 2023-05-11
Fix circular ref

DavieReid:


docs/configure/modes/index.mdx
```diff
@@ -5,7 +5,7 @@ sidebar:
   priority: 4
 data:
   modes:
-    $ref: ./*#/title
```
DavieReid, line 8: title is also a property of this page creating the circular ref.

PR #356, opened 2023-05-11
Dynamically add a source

DavieReid:
This allows you to push a source definition to an active mosaic instance.  Do we think it should add this definition to the mosaic config file as well?

## Questions/Issues

- At the moment, you could push a source definition and then if there is a restart the definition would be lost.
~~Also, disabled this in prod env.~~
- local folder definitions probably won't work, because the local folder needs to exist where mosaic is running

### Example

Post request to http://localhost:8080/sources/add

request body:

```
{
    "name":"davie-docs-preview",
    "definition": {
      "modulePath": "@jpmorganchase/mosaic-source-git-repo",
      "namespace": "davie",
      "options": {
        "credentials":<access-token> ,
        "prefixDir": "davie",
        "cache": true,
        "subfolder": "docs",
        "repo": "https://github.com/DavieReid/mosaic-docs.git",
        "branch": "main",
        "extensions": [".mdx"],
        "remote": "origin"
      }
    }
}
```

### DIFF
--- a/packages/cli/src/serve.ts
+++ b/packages/cli/src/serve.ts
@@ -80,6 +80,10 @@ export default async function serve(config, port, scope) {
     try {
       const { definition, name, isPreview = true } = req.body;
 
+      if (process.env.MOSAIC_ENABLE_SOURCE_PUSH !== 'true') {
+        throw new Error('Source push is disabled.');
+      }
+
       if (!definition) {
         throw new Error('Source definition is required');
       }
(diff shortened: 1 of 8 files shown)


packages/cli/src/serve.ts
```diff
@@ -73,4 +73,29 @@ export default async function serve(config, port, scope) {
       res.status(500).send(e.message).end();
     }
   });
+
+  app.post('/sources/add', async (req, res) => {
+    try {
+      const { definition, isPreview = true } = req.body;
+
+      if (process.env.NODE_ENV === 'production') {
+        throw new Error('Sources cannot be added in production.');
```
mark-tate, line 82: nice catch

joshwooding (approved):

lilyvc (approved):

PR #358, opened 2023-05-17
feat: update font loading to use `next/font`

DavieReid:


packages/site/src/pages/_app.tsx
```diff
@@ -1,18 +1,35 @@
...
 import { SessionProvider } from 'next-auth/react';
+import classnames from 'clsx';
 
 import { MyAppProps } from '../types/mosaic';
 
+const ptMono = PT_Mono({
+  weight: '400',
+  display: 'swap',
+  subsets: ['latin'],
+  variable: '--salt-typography-fontFamily:'
```
DavieReid, line 25: Not sure if we need another variable name thats not salt specific but this works for now

packages/site/e2e/layout.detailHighlight.test.ts
```diff
@@ -1,9 +1,10 @@
 import { test, expect } from '@playwright/test';
 
+import { waitForFonts } from './utils';
+
 test('take a screenshot', async ({ page }) => {
   await page.goto('/mosaic/test/layouts/detail-highlight');
-  await page.evaluate(() => document.fonts.ready);
-  await page.waitForFunction(() => document.fonts.check('18px Open Sans'));
+  await page.waitForFunction(waitForFonts);
```
mark-tate, line 7: would globalSetup work for this ?

mark-tate (approved):

packages/site/e2e/layout.detailHighlight.test.ts
DavieReid: might do, but can't get the tests to pass at the moment.  Fonts aren't loaded correctly in the test

PR #360, opened 2023-05-19
Improve generator

mark-tate:
As a Mosaic Site creator we previously installed `@jpmorganchase/mosaic-create-site` globally and run

`yarn mosaic-create-site create -o my-sample-site`

We have switched the docs to use `npx` instead to avoid the need to globally install.

When running the create, we would read `mosaic.generators.js` which read the `@jpmorganchase/mosaic-standard-generator` using `require`. The problem was that for this to work, it needed a `package.json`. This worked OK, in monorepo situations where you have a `package.json` anyway but not for creating new projects from scratch.

By switching to `npx` we need to add the temporary path of `node-modules`, so that it can require in the standard generator without needing a local package.json. Owing to the way that require searches for references, this change will work for both monorepo and new project situations.

### DIFF
--- a/docs/quick-start/publish-site-to-aws.mdx
+++ b/docs/quick-start/publish-site-to-aws.mdx
@@ -10,8 +10,7 @@ Publish a site to AWS using S3 snapshots.
 ## Step 1: Generate a Mosaic site
 
 ```
-> yarn global add @jpmorganchase/mosaic-create-site
-> mosaic-create-site create -o my-sample-site
+> npx @jpmorganchase/mosaic-create-site -o my-sample-site
 > cd my-sample-site
 ```
 
--- a/docs/quick-start/site-generator.mdx
+++ b/docs/quick-start/site-generator.mdx
@@ -17,9 +17,7 @@ Use the Mosaic Site generator to generate a site and add an example source.
 The site generator expects to be run in a monorepo. If you are trying to set up a site in an empty directory, you will first need to run `yarn init` to create a `package.json` file.
 
 ```
-> yarn add @jpmorganchase/mosaic-create-site -W
-> yarn mosaic-create-site init
-> yarn mosaic-create-site create -o my-sample-site
+> npx @jpmorganchase/mosaic-create-site -o my-sample-site
 > cd my-sample-site
 ```
 
--- a/packages/create-site/README.md
+++ b/packages/create-site/README.md
@@ -57,7 +57,7 @@ Once you have created your `mosaic.generators.js`, we can think about configurin
 Running `create` will generate a Mosaic site from the configured templates.
 Lets see what this looks like by running the script with the `--interactive (-i)` and the `--output (-o)` flags.
 
-`npx @jpmorganchase/mosaic-create-site  --interactive --ouput <path to your output directory>`
+`npx @jpmorganchase/mosaic-create-site --interactive --ouput <path to your output directory>`
 
 You should see a menu appear giving you a list of the available Mosaic templates.
 
--- a/packages/create-site/README.md
+++ b/packages/create-site/README.md
@@ -2,13 +2,9 @@
 
 `@jpmorganchase/mosaic-create-site` is a CLI script which scaffolds your very own Mosaic site from pre-defined generators.
 
-## Installation
-
-`yarn add -g @jpmorganchase/mosaic-create-site`
-
 ## Basic Commands
 
-### `yarn mosaic-create-site init`
+### `npx @jpmorganchase/mosaic-create-site init`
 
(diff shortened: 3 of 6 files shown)


packages/standard-generator/src/generator.config.js
```diff
@@ -74,6 +74,7 @@ module.exports = {
        */
       modulePath: '@jpmorganchase/mosaic-source-git-repo',
       namespace: 'mosaic', // each site has it's own namespace, think of this as your content's uid
+      disabled: !process.env.MOSAIC_DOCS_CLONE_CREDENTIALS, // use a boolean to enable/disable the source
```
mark-tate, line 77: The simplest possible site, is starting up with just local docs.
You cannot start with that, as Mosaic will fail to start if `MOSAIC_DOCS_CLONE_CREDENTIALS` is undefined

Use this to toggle the boolean, so that we can progressively enhance the local setup to load local docs and then both local and remote docs.

As the initial startup is the welcome mat for new users, we should not start their experience with an error, which they may not yet understand.

JoanaMMoreira (approved):

docs/quick-start/site-generator.mdx
```diff
@@ -17,9 +17,7 @@ Use the Mosaic Site generator to generate a site and add an example source.
 The site generator expects to be run in a monorepo. If you are trying to set up a site in an empty directory, you will first need to run `yarn init` to create a `package.json` file.
 
 ```
-> yarn add @jpmorganchase/mosaic-create-site -W
-> yarn mosaic-create-site init
-> yarn mosaic-create-site create -o my-sample-site
+> npx @jpmorganchase/mosaic-create-site -o my-sample-site
```
lilyvc, line 20: is it not still mosaic-create-site *create* -o my-sample-site? 
I can't see any changes that mean we can remove create

packages/create-site/README.md
```diff
@@ -2,13 +2,9 @@
...
 `@jpmorganchase/mosaic-create-site` is a CLI script which scaffolds your very own Mosaic site from pre-defined generators.
 
-## Installation
-
-`yarn add -g @jpmorganchase/mosaic-create-site`
-
 ## Basic Commands
 
-### `yarn mosaic-create-site init`
+### `npx @jpmorganchase/mosaic-create-site init`
```
lilyvc, line 7: I feel like `init` implies it is something that is required before you can create a site. I think having it here under basic commands might be a bit confusing for the reader. 

not necessarily in the scope of this PR

lilyvc (approved):

docs/quick-start/publish-site-to-aws.mdx
```diff
@@ -10,8 +10,7 @@ Publish a site to AWS using S3 snapshots.
 ## Step 1: Generate a Mosaic site
 
 ```
-> yarn global add @jpmorganchase/mosaic-create-site
-> mosaic-create-site create -o my-sample-site
+> npx @jpmorganchase/mosaic-create-site -o my-sample-site
```
lilyvc, line 13: I tried to do this in a GitHub codespace and got this error:

$ npx @jpmorganchase/mosaic-create-site create -o my-sample-site
node:internal/modules/cjs/loader:1090
  const err = new Error(message);
              ^

Error: Cannot find module '/usr/local/share/nvm/versions/node/v19.9.0/lib/node_modules/@jpmorganchase/mosaic-standard-generator/dist/generator.js'
Require stack:
- /home/codespace/.npm/_npx/42ed83063eb32b24/node_modules/resolve-global/index.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1090:15)
    at Function.resolve (node:internal/modules/helpers:125:19)
    at resolveGlobal (/home/codespace/.npm/_npx/42ed83063eb32b24/node_modules/resolve-global/index.js:9:18)
    at resolvePath (file:///home/codespace/.npm/_npx/42ed83063eb32b24/node_modules/@jpmorganchase/mosaic-create-site/dist/create.mjs:37:20)
    at file:///home/codespace/.npm/_npx/42ed83063eb32b24/node_modules/@jpmorganchase/mosaic-create-site/dist/create.mjs:47:27
    at Array.map (<anonymous>)
    at createMosaicApp (file:///home/codespace/.npm/_npx/42ed83063eb32b24/node_modules/@jpmorganchase/mosaic-create-site/dist/create.mjs:45:32) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/codespace/.npm/_npx/42ed83063eb32b24/node_modules/resolve-global/index.js'
  ]
}

whereas doing it the old way worked (yarn global add @jpmorganchase/mosaic-create-site
then mosaic-create-site create -o my-sample-site)

maybe the npx way isn't as reliable or has more prequisites, could we suggest both? @mark-tate @DavieReid

PR #364, opened 2023-05-25
Component props table plugin 

JoanaMMoreira:
This is a plugin that generates a component props table given a file path e.g. `:propsTable{src="../../packages/core/src/button/Button.tsx"}` 

It will allow users to embed the table within their mdx pages.

Screenshot 2023-05-24 at 16 05 28


yarn.lock
```diff
@@ -11286,6 +11286,11 @@ react-debounce-input@=3.3.0:
     lodash.debounce "^4"
     prop-types "^15.8.1"
 
+react-docgen-typescript@^2.2.2:
+  version "2.2.2"
+  resolved "https://artifacts-read.gkp.jpmchase.net/artifactory/api/npm/npm/react-docgen-typescript/-/react-docgen-typescript-2.2.2.tgz#4611055e569edc071204aadb20e1c93e1ab1659c"
```
lilyvc, line 11291: the checks are failing because it's pointing at the internal repository

DavieReid:
Committed search data in the public dir. that's not needed

lilyvc (approved):

mark-tate (approved):

mark-tate:
how does the path work to the component work ?
Will this only work when run locally in a monorepo that contains the components ?

PR #370, opened 2023-05-31
Fix styles audio player

lukeac123:
Fix styles to centre the button bar

### DIFF
--- a/packages/components/src/AudioPlayer/index.tsx
+++ b/packages/components/src/AudioPlayer/index.tsx
@@ -132,12 +132,7 @@ export const AudioPlayer: React.FC<AudioPlayerProps> = ({ src, title, skipDurati
         <Caption6>{durationString}</Caption6>
       </div>
 
-      <ButtonBar
-        className={styles.buttonBar}
-        stackAtBreakpoint={0}
-        // alignLeft={false}
-        disableAutoAlignment={true}
-      >
+      <ButtonBar className={styles.buttonBar} stackAtBreakpoint={0} disableAutoAlignment={true}>
         <div>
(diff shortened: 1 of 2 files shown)


packages/components/src/AudioPlayer/styles.css.ts
```diff
@@ -52,7 +52,7 @@ export default {
   ]),
   buttonBar: style({
     position: 'relative',
-    justifyContent: 'center',
+    justifyContent: 'center !important',
```
lukeac123, line 55: Salt styles were taking precedence, even when setting this in the vanilla extract theme.
!important flag used to make sure it is always applied

packages/components/src/AudioPlayer/index.tsx
```diff
@@ -132,7 +132,7 @@ export const AudioPlayer: React.FC<AudioPlayerProps> = ({ src, title, skipDurati
         <Caption6>{durationString}</Caption6>
       </div>
 
-      <ButtonBar className={styles.buttonBar} stackAtBreakpoint={0}>
+      <ButtonBar className={styles.buttonBar} stackAtBreakpoint={0} disableAutoAlignment={true}>
```
tomhazledine, line 135:
```suggestion
      <ButtonBar className={styles.buttonBar} stackAtBreakpoint={0} disableAutoAlignment>
```

Value must be omitted for boolean attributes: [eslint react/jsx-boolean-value](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/jsx-boolean-value.md)

tomhazledine (approved):
Not wild about adding `!important` but don't think it should hold up releasing a fix.

PR #371, opened 2023-05-31
Refactor TOC

joshwooding:


lukeac123:
Just had a play around with it in preview, the TOC seems to be changing in accordance with the bottom of the page instead of the top. I assume this is due to the use of intersection observer api and when the respective content comes into the viewport. Can this be modified so the TOC changes to the section when the content is at the top of the viewport ?

joshwooding:
> Just had a play around with it in preview, the TOC seems to be changing in accordance with the bottom of the page instead of the top. I assume this is due to the use of intersection observer api and when the respective content comes into the viewport. Can this be modified so the TOC changes to the section when the content is at the top of the viewport ?

Changing in accordance to the top can cause issues where if the last section is too short the TOC will not highlight the last item but I might be able to balance it some more.

PR #378, opened 2023-06-07
App router prep

mark-tate:
This 2 commits, upgrade esbuild and Next  in prep for any future migration to RSCs

esbuild: now preserves directives
nextjs: resolves an issue importing RSCs from our components/site-components package

We still use the page router, so this PR is backwardly compatible with the existing code.
The intention of this PR is to reduce the changes for any future PR that adds RSC support


lilyvc (approved):

DavieReid:
Should er hold off on this for now until the Developer site work is completed?  Or alternatively we could create a next branch and set up another deployment for this?

PR #381, opened 2023-06-08
DO NOT MERGE 324 replace prism js with shiki

DavieReid:
Hi @origami-z

### DIFF
--- a/packages/site-middleware/src/plugins/prettyCode.ts
+++ b/packages/site-middleware/src/plugins/prettyCode.ts
@@ -1,4 +1,11 @@
+import { getHighlighter } from 'shiki';
+
 const options = {
+  getHighlighter: highlighterOptions =>
+    getHighlighter({
+      ...highlighterOptions,
+      langs: ['bat', 'diff', 'css', 'json', 'jsx', 'tsx', 'mdx', 'yaml', 'ts']
+    }),
   theme: {
     dark: 'dark-plus',
     light: 'light-plus'
--- a/packages/site/scripts/copy-shiki.mjs
+++ b/packages/site/scripts/copy-shiki.mjs
@@ -25,12 +25,14 @@ try {
   console.log(`from: ${shikiModuleDir}`);
   console.log(`to: ${publicShikiDir}`);
   console.groupEnd();
+
   await fs.promises.cp(
-    path.posix.join(shikiModuleDir, 'themes'),
-    path.posix.join(publicShikiDir, 'themes'),
-    {
-      recursive: true
-    }
+    path.posix.join(shikiModuleDir, 'themes', 'light-plus.json'),
+    path.posix.join(publicShikiDir, 'themes', 'light-plus.json')
+  );
+  await fs.promises.cp(
+    path.posix.join(shikiModuleDir, 'themes', 'dark-plus.json'),
+    path.posix.join(publicShikiDir, 'themes', 'dark-plus.json')
   );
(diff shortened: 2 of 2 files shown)


DavieReid:
> Depending on how long it takes to get Shiki to work, it would be great to have the commits where you added the languages to the docs in a separate PR. The code changes will still be highlighted by Prism so the site will look better.

I don't think the way languages are specified is the same...is it?

packages/site/next.config.js
```diff
@@ -36,7 +38,15 @@ module.exports = {
     // Swaps out Buble for a smaller version that removes the latest Regex spec features.
     // See https://github.com/FormidableLabs/react-live#what-bundle-size-can-i-expect
     config.plugins.push(
-      new webpack.NormalModuleReplacementPlugin(/^buble$/, require.resolve('@philpl/buble'))
+      new webpack.NormalModuleReplacementPlugin(/^buble$/, require.resolve('@philpl/buble')),
+      new CopyPlugin({
+        patterns: [
+          {
+            from: path.posix.resolve(path.posix.dirname(require.resolve('shiki')), '..'),
```
mark-tate, line 45: potentially another windows issue

I'm assuming we are going to go down the container route for local development.

mark-tate (approved):

lilyvc:
> > Depending on how long it takes to get Shiki to work, it would be great to have the commits where you added the languages to the docs in a separate PR. The code changes will still be highlighted by Prism so the site will look better.
> 
> I don't think the way languages are specified is the same...is it?

oh you're right there is some extra syntax that would break our pages..
some of it would work like ```js but yea not worth separating

packages/site-middleware/src/plugins/prettyCode.ts
```diff
@@ -1,4 +1,11 @@
+import { getHighlighter } from 'shiki';
+
 const options = {
+  getHighlighter: highlighterOptions =>
+    getHighlighter({
+      ...highlighterOptions,
+      langs: ['bat', 'diff', 'css', 'json', 'jsx', 'tsx', 'mdx', 'yaml', 'ts']
```
DavieReid, line 7: I don't think we use any other languages in code samples on the site...

@lilyvc

lilyvc: maybe js? for a config.js

packages/site-middleware/src/plugins/prettyCode.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const options = {
+  getHighlighter: highlighterOptions =>
+    getHighlighter({
+      ...highlighterOptions,
+      langs: ['bat', 'diff', 'css', 'json', 'jsx', 'tsx', 'mdx', 'yaml', 'ts']
+    }),
+  theme: {
+    dark: 'dark-plus',
+    light: 'light-plus'
```
lilyvc, line 11: I don't love how the light theme looks, is it easy to switch it out?

packages/site-middleware/src/plugins/prettyCode.ts
DavieReid: Good shout

packages/site-middleware/src/plugins/prettyCode.ts
DavieReid: Very easy to change. I'm not a huge fan either but picked this with a11y in mind.

PR #382, opened 2023-06-12
Update Salt packages

joshwooding:
This PR upgrades Salt packages


packages/theme/scripts/saltIconNames.js
```diff
@@ -17,7 +17,7 @@ const saltIconNames = {
   name: 'iconEnum',
   setup(build) {
     const icons = [];
-    build.onLoad({ filter: /src[\/\\]components/ }, async args => {
+    build.onLoad({ filter: /components/ }, async args => {
```
joshwooding, line 20: Due to the new style injection the internal structure of the package has changed

patches/@salt-ds+lab+1.0.0-alpha.9.patch
joshwooding, line 1: This is more a test but this shouldn't be needed now 🤞🏽

tomhazledine (approved):
LGTM 👍 

As the `StatusBanner` is in the content editor (and thus not easily tested in the example site) we should keep an eye on it in future, but I'm happy with the code changes

Seeing an issue locally where fonts aren't being loaded with the correct specificity, but after digging into it am pretty sure it's unrelated to this PR

DavieReid (approved):

DavieReid:
> LGTM 👍
> 
> As the `StatusBanner` is in the content editor (and thus not easily tested in the example site) we should keep an eye on it in future, but I'm happy with the code changes
> 
> Seeing an issue locally where fonts aren't being loaded with the correct specificity, but after digging into it am pretty sure it's unrelated to this PR

Lily fixed a mistake I made with the fonts the other day...check you have the latest code

PR #385, opened 2023-06-16
Content editor

DavieReid:
Updates for the Content Editor aimed at ensuring design and functionality matches what we have previously built/used


packages/content-editor-plugin/src/components/Dialog.tsx
lilyvc, line 1: why use this instead of just the Salt Dialog?

DavieReid: The Salt Dialog alone is in a portal that is outside the scope of the mosaic theme. So this component is essentially just a wrapper around the salt dialog that also applies the mosaic theme class name so our styles work. 

We could potentially move where we apply the theme class name (apply to the body rather than theme provide) but didn't know what else that might break

PR #389, opened 2023-06-19
use double quotes so serve command works on windows

lilyvc:
Was running the Mosaic site on windows, and realised this command needs to be updated. Also updated it in the template so newly generated sites support windows.


packages/standard-generator/src/templates/package.json.hbs
```diff
@@ -16,7 +16,7 @@
     "gen:snapshot": "yarn mosaic build --out snapshots --name latest --config mosaic.config.js",
     "serve:snapshot:file": "yarn cross-env MOSAIC_MODE='snapshot-file' concurrently --kill-others 'yarn dev'",
     "serve:snapshot:s3": "yarn cross-env MOSAIC_MODE='snapshot-s3' concurrently --kill-others 'yarn dev'",
-    "serve": "concurrently --kill-others 'yarn dev' 'yarn mosaic serve -c ./mosaic.config.js' -p 8080"
+    "serve": "concurrently --kill-others \"yarn dev\" \"yarn mosaic serve -c ./mosaic.config.js\" -p 8080"
```
libertymayc, line 19: should double quotes be on the ones above too?

mark-tate: good shout @libertymayc

DavieReid (approved):

PR #394, opened 2023-06-26
Manually dedupe @aws-sdk/client-s3

joshwooding:
Follow up #393 that dedupes the yarn.lock


packages/site-middleware/src/__tests__/withMDXContent.test.ts
```diff
@@ -1,6 +1,6 @@
 import { GetObjectCommand, S3Client } from '@aws-sdk/client-s3';
 import { AwsStub, mockClient } from 'aws-sdk-client-mock';
-import { sdkStreamMixin } from '@aws-sdk/util-stream-node';
+import { sdkStreamMixin } from '@aws-sdk/util-stream';
```
joshwooding, line 3: the client sdk doesn't rely on util-stream-node directly now

DavieReid (approved):

PR #397, opened 2023-06-29
Ibce preview

DavieReid:
Refactors required for In browser content editor to become functional.

Another key change is that `mosaic.config.js` has been renamed everywhere to `mosaic.config.mjs`.  This allows us to import workflows into the config file but also is consistent with our approach of trying to have "ESM everywhere".

Hi @origami-z


packages/site/mosaic.config.mjs
```diff
@@ -53,6 +45,20 @@ module.exports = deepmerge(siteConfig, {
         prefixDir: 'mosaic', // root path used for namespace
         extensions: ['.mdx'] // extensions of content which should be pulled
       }
+    },
```
DavieReid, line 48: oops will remove

packages/site/src/pages/api/content/preview.ts
```diff
@@ -0,0 +1,26 @@
+import { compileMDX } from '@jpmorganchase/mosaic-site-middleware';
```
DavieReid, line 1: Not sure if this should be removed (as we don't need it) or if it should be kept and also added to the site generator template...

This the endpoint the preview component hits to compile the markdown pulled from the IBCE

thoughts? @lukeac123 @lilyvc @mark-tate @tomhazledine

packages/site/mosaic.config.mjs
DavieReid: DONE

packages/site/.env.local
```diff
@@ -3,4 +3,7 @@ MOSAIC_SNAPSHOT_DIR=snapshots/latest
 OPTIMIZE_IMAGES=false
 NEXTAUTH_URL=http://localhost:3000
 NODE_ENV=development
-MOSAIC_ENABLE_SOURCE_PUSH=true
\ No newline at end of file
+MOSAIC_ENABLE_SOURCE_PUSH=true
+NEXTAUTH_URL=http://localhost:3000/
```
libertymayc, line 7: repeat of above? or you need one with a slash?

packages/site/src/pages/api/content/preview.ts
lukeac123: If it's not being used what's the reasoning to keep it ?

packages/site/.env.local
DavieReid: no copy & paste error...will remove in next PR

lukeac123: think it works with either

lukeac123 (approved):

lilyvc (approved):

PR #401, opened 2023-06-30
fix: ensure main content area uses all available height

DavieReid:
When there is not enough content to fill the page, dark mode looks awful as half the page is white.

See here --> https://mosaic-mosaic-dev-team.vercel.app/mosaic/author/index

Also fixes the label color of the active sidebar item when in dark mode.


JoanaMMoreira:
Should the footer also stick to the bottom?

DavieReid:
> Should the footer also stick to the bottom?

Not with the current design...next design will have a full-width footer than sticks to the bottom of the page

JoanaMMoreira (approved):

PR #403, opened 2023-07-03
Bump swagger-ui-react to ^5.0.0

joshwooding:
Helps close https://github.com/jpmorganchase/mosaic/security/dependabot/18

I believe swagger-ui-react is the only shipped dependency that is affected


mark-tate (approved):
with the app router changes we will remove this component.. folks can add it themselves, if they want it.

PR #407, opened 2023-07-11
fix: remove use of lodash `merge` from `$RefPlugin`

DavieReid:
When a page that is referenced via a wildcard ref **is updated**, the wildcard ref can be written into that page when the target page uses the same path as the referenced page.  This happens in the `$beforeSend` lifecycle event of the `$RefPlugin`

### Example

Page A with a wildcard ref
```
data:
  items:
    $ref: ./*#/data
```

Page B (a target)
```
data
  title: 'a title'
```

When the wildcard ref of page A is resolved it can result in the metadata for Page B being incorrectly updated to

```
data
  title: 'a title'
  $ref:
       'path-to-page-B#/data'
```

This is a circular ref because Page B now has a ref that points to its own data property.

### DIFF
--- a/.changeset/shiny-planes-deliver.md
+++ b/.changeset/shiny-planes-deliver.md
@@ -1,0 +1,6 @@
+---
+'@jpmorganchase/mosaic-plugins': patch
+'@jpmorganchase/mosaic-site': patch
+---
+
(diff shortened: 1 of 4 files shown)


tambo (approved):

tambo:
looks good

PR #412, opened 2023-07-13
Getting started

DavieReid:
Updates to ensure site generator works and is windows compatible

### DIFF
--- a/packages/standard-generator/src/templates/package.json.hbs
+++ b/packages/standard-generator/src/templates/package.json.hbs
@@ -15,7 +15,7 @@
     "start": "next start",
     "gen:snapshot": "yarn mosaic build --out snapshots --name latest --config mosaic.config.mjs",
     "serve:snapshot:file": "yarn cross-env MOSAIC_MODE=\"snapshot-file\" concurrently --kill-others \"yarn dev\"",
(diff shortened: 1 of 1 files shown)


packages/create-site/bin/mosaic-create-site.js
```diff
@@ -43,9 +43,6 @@ async function readMosaicConfig(mosaicConfigPath) {
     await init({ silent: true });
   }
   try {
-    const nodeModulesPathRegexp = /(.*\/node_modules\/)/;
-    const nodeModulesPathMatches = __filename.match(nodeModulesPathRegexp);
-    process.env.NODE_PATH = `${process.env.NODE_PATH}${path.delimiter}${nodeModulesPathMatches[1]}`;
```
DavieReid, line 48: @mark-tate @joshwooding 

I don't know why we are doing this...can you shed any light?

It breaks the site generator on windows and removing seems to get everything working on mac and windows

mark-tate (approved):

packages/create-site/bin/mosaic-create-site.js
origami-z: I quickly tried commenting these line in npx install location, it will not be able to find mosaicConfig

PR #418, opened 2023-07-20
Tile design fixes

DavieReid:
Address styling issues with Mosaic Tile component(s)

### DIFF
--- a/packages/components/src/LinkText/styles.css.ts
+++ b/packages/components/src/LinkText/styles.css.ts
@@ -1,5 +1,17 @@
 import { link, responsiveSprinkles } from '@jpmorganchase/mosaic-theme';
-import { style } from '@vanilla-extract/css';
+import { style, globalStyle } from '@vanilla-extract/css';
+
+const iconStyle = style([
+  {
+    color: 'inherit',
+    stroke: 'currentColor',
+    height: '1em'
+  },
+  responsiveSprinkles({
+    paddingX: ['x1', 'x1', 'x1', 'x1']
+  })
+]);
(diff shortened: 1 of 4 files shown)


packages/components/src/LinkText/styles.css.ts
```diff
@@ -9,13 +21,10 @@ export default {
...
-    },
-    responsiveSprinkles({
-      paddingX: ['x1', 'x1', 'x1', 'x1']
-    })
-  ])
+  icon: iconStyle
 };
+
+globalStyle(`${iconStyle} > svg.saltIcon`, {
+  height: 'calc(1em - 3px)',
```
DavieReid, line 28: The icon is an svg now so we can't style it via fontSize anymore.

packages/plugins/src/$RefPlugin.ts
```diff
@@ -139,7 +140,7 @@ const $RefPlugin: PluginType<RefsPluginPage, unknown, RefsPluginConfigData> = {
           pageExtensions,
           ignorePages
         );
-        normalisedRefs[fullPath] = { ...page, ...expandedRefs };
+        normalisedRefs[fullPath] = deepmerge(page, expandedRefs);
```
DavieReid, line 143: Without this, we were missing info that may be included inside the `data` metadata of a page BUT is not part of a ref.  The merge here seems to fix the issue of dropping non-ref'd data while still preventing the circular ref issue encountered when we were using lodash merge.  Will continue to monitor the behaviour of the ref plugin.

PR #426, opened 2023-07-27
Investigate toc

DavieReid:


packages/site-components/src/TableOfContents/useIntersectionObserver.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      if (visibleHeadings.length === 0) {
+        return;
+      }
+
+      selectActive(visibleHeadings.map(entry => entry.target.id));
+    };
+
+    const observer = new IntersectionObserver(handleIntersection, {
+      rootMargin: '-30% 0px -30% 0px',
```
mark-tate, line 36: these feel a little magical, do they need to be configurable ?

mark-tate (approved):

mark-tate:
it works better, lets get it in and we can iterate as we need to

DavieReid:
Another look at this and the implementation is likely to change again with the move to app directory

PR #430, opened 2023-07-31
429 source retry

DavieReid:
Source Retries

The source retries feature gracefully handles re-requesting content from a source that has thrown an error. Retries follow an exponential back-off strategy and will eventually rethrow the error when all retry attempts have been exhausted.

The retry configuration is specified as part of the source schedule e.g.

```
  schedule: {
    checkIntervalMins: 0.2,
    initialDelayMs: 2000,
    retryDelayMins: 5,
    maxRetries: 30,
    retryEnabled: true
  },
```


lilyvc (approved):

packages/core/src/worker/helpers/createSourceObservable.ts
```diff
@@ -74,6 +75,13 @@ NOTE: Only ${pageExtensions.join(
           route: page.route ? page.route.toLowerCase() : page.fullPath.toLowerCase()
         });
       }, [])
-    )
+    ),
+    exponentialBackOffRetryStrategy(schedule),
+    catchError(error => {
+      console.log(
+        `[Mosaic] Source failed and retries are ${schedule.retryEnabled ? 'exhausted' : 'disabled'}`
```
mark-tate, line 82: will this identify which source became disabled ?

packages/core/src/worker/helpers/exponentialBackOffRetryStrategy.ts
```diff
@@ -0,0 +1,26 @@
...
+  (schedule: SourceSchedule) => (attempts: Observable<Page[]>) =>
+    attempts.pipe(
+      retry({
+        count: schedule.retryEnabled ? schedule.maxRetries : 0,
+        delay: (_error, retryIndex) => {
+          const retryDelayMs = schedule.retryDelayMins * 60000;
+          // eslint-disable-next-line prefer-exponentiation-operator, no-restricted-properties
+          const backOffTime = Math.pow(2, retryIndex - 1) * retryDelayMs;
+          const remainingAttempts = schedule.maxRetries - retryIndex;
+          console.warn('[Mosaic] Retrying failed source');
```
mark-tate, line 15: does this need to identify the source

packages/schemas/src/SourceScheduleSchema.ts
```diff
@@ -8,7 +8,27 @@ export const sourceScheduleSchema = z.object({
...
+   * @default 5 mins
+   */
+  retryDelayMins: z.number().optional().default(5),
+  /**
+   *  Maximum number of retry attempts
+   * @default 100
+   */
+  maxRetries: z.number().optional().default(100),
+  /**
+   * If true, after a successful retry the retry logic is reset to its initial settings
```
mark-tate, line 28: NIT comment was a little unclear until I looked up the RXJS api

I think this means the retry counter is reset to zero once it recovers.

mark-tate: What would the use-case where you would not resetOnSuccess ?

packages/core/src/worker/helpers/exponentialBackOffRetryStrategy.ts
DavieReid: Can't here. The id is a generated symbol and can't be passed to the source worker thread via worker data as it's incompatible with structured cloning.

DavieReid: We actually have a `name` property that isn't a symbol being passed to the worker.  Pushed a commit that uses that:

Example log:

```
[Mosaic] Source 'site#91fb4fd4' started in 1386ms - awaiting first docs snapshot
[Mosaic] Source 'site#91fb4fd4' received first docs snapshot 0.001s after starting.
[Mosaic] Retrying failed source: site#91fb4fd4
[Mosaic] Attempt (1/2). 1 attempt remaining
[Mosaic] Retry delay: 1 mins for source: site#91fb4fd4
```

packages/core/src/worker/helpers/createSourceObservable.ts
DavieReid: Pushed another commit:

Example log:

```
[Mosaic] Source site#91fb4fd4 failed and retries are exhausted
```

packages/schemas/src/SourceScheduleSchema.ts
DavieReid: Updated the comment.  I think we would always want to reset it to true, but we have a variety of sources so wanted to be as flexible as possible.

mark-tate (approved):

mark-tate (approved):

PR #431, opened 2023-08-02
Update sitemap component

maleeha13:
- Retrieves the routes from search-data-condensed and uses that in the sitemap component
- Sitemap component works with new version of mosaic so can be used in go/developer when version is updated
- Creates new page for sitemap component


docs/sitemap/index.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: Sitemap
+layout: DetailTechnical
+sidebar:
+
+data:
+  exampleRefData: Hello from Sitemap page
+---
+
+<Labs.Sitemap href="http://localhost:3000/api/data/data" />
```
DavieReid, line 10: please change the API so that its purpose is known. `api/data/data` could be anything

docs/sitemap/index.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: Sitemap
+layout: DetailTechnical
+sidebar:
+
+data:
+  exampleRefData: Hello from Sitemap page
```
DavieReid, line 7: Why?

packages/site/src/pages/api/data/data.js
```diff
@@ -0,0 +1,35 @@
+const fs = require('fs');
```
DavieReid, line 1: You could have modified the sitemap plugin that we already have to do this I think.  We don't need to do it this way

DavieReid:
This couples the site map component to search data....2 unrelated things.

The preview does not have a sitemap page...

DavieReid:
No action for 2 weeks

PR #432, opened 2023-08-03
fix: reintroduce support for tab `label`

DavieReid:
Salt switched from `label` to `title` for Tabs but we can support both to prevent the need for content changes.


libertymayc (approved):

libertymayc:
does this need a changeset?

joshwooding (approved):

DavieReid:
> does this need a changeset?

yes!

lilyvc (approved):

Issue #447, opened 2023-08-11
Swagger CSS not included in OpenApiComponent styles


DavieReid:
The OpenApiComponent renders the react-swagger component without any styling.

## Root Cause
I suspect the root cause is the fact that esbuild is told to exclude node_modules when bundling the open API component
https://github.com/jpmorganchase/mosaic/blob/main/scripts/bundle.js#L36

## Workaround
Add this to `_app.tsx`

```
import 'swagger-ui-react/swagger-ui.css';
```

MayankPrasher:
Still open ?

DavieReid:
> Still open ?

Yes, still open

PR #448, opened 2023-08-11
Update packages/site/src/pages/_app.tsx

viktoriussuwandi:
fix issue #447

Hi [DavieReid](https://github.com/DavieReid),

Just trying to jump on this issue, please check it out

### DIFF
--- a/packages/site/src/pages/_app.tsx
+++ b/packages/site/src/pages/_app.tsx
@@ -17,6 +17,7 @@ import { SessionProvider } from 'next-auth/react';
 import classnames from 'clsx';
 
 import { MyAppProps } from '../types/mosaic';
+import 'swagger-ui-react/swagger-ui.css';
 
 const ptMono = PT_Mono({
   weight: '400',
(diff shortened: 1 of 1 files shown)


DavieReid:
Hi, thanks for picking this up so quickly!

However, this is just applying the workaround mentioned in the issue, and we don't use the open api component on the mosaic site at the moment anyway.

I would prefer that we took a look at the build of the open api component so that it's css file includes the swagger css.  You can see it's included here --> https://github.com/jpmorganchase/mosaic/blob/main/packages/open-api-component/src/styles.ts

If you build the component and then look in the `dist` folder at the styles.css file you will only see the CSS from here --> https://github.com/jpmorganchase/mosaic/blob/main/packages/open-api-component/src/OpenAPI/styles.css.ts

Ideally, the following CSS import would correctly style the swagger component:

```
import '@jpmorganchase/mosaic-open-api-component/index.css';
```

viktoriussuwandi:
Thanks for your suggestion @DavieReid, already update import statement as mentioned

DavieReid:
> Thanks for your suggestion @DavieReid, already update import statement as mentioned

This is not what I suggested.  Your fix does not address the root cause of the problem.

DavieReid:
Closing as doesn't address the problem.

PR #450, opened 2023-08-14
Admin urls

DavieReid:
Provide a few APIs to expose info about the mosaic fs and manage active sources


packages/cli/src/serve.ts
```diff
@@ -22,6 +21,64 @@ export default async function serve(config, port, scope) {
...
+  app.put('/_mosaic_/source/stop', async (req, res) => {
+    res.contentType('application/text');
+    const { name } = req.body;
+    if (name) {
+      try {
+        await mosaic.stopSource(String(name));
+        res.send(`${name} stopped successfully.`);
+      } catch (e) {
+        console.error(e);
+        res.status(500).send(`${name} not stopped.  Check you have the right name!`).end();
```
joshwooding, line 60:
```suggestion
        res.status(500).send(`${name} not stopped.  Check you have the right name!`);
```

packages/cli/src/serve.ts
```diff
@@ -22,6 +21,64 @@ export default async function serve(config, port, scope) {
...
+  app.put('/_mosaic_/source/restart', async (req, res) => {
+    res.contentType('application/text');
+    const { name } = req.body;
+    if (name) {
+      try {
+        await mosaic.restartSource(String(name));
+        res.send(`${name} restarted successfully.`);
+      } catch (e) {
+        console.error(e);
+        res.status(500).send(`${name} not restarted.  Check you have the right name!`).end();
```
joshwooding, line 77:
```suggestion
        res.status(500).send(`${name} not restarted.  Check you have the right name!`);
```

joshwooding (approved):

packages/cli/src/serve.ts
DavieReid: Will follow up in another PR

packages/cli/src/serve.ts
DavieReid: Will follow up in another PR

PR #452, opened 2023-08-17
Shared config updates

DavieReid:
It is possible to have a source that is considered a "child" of another source.  Consider:

- Source A contains a lot of product docs and defines a shared config for the products site (parent source)
- Source B contains docs for a single product and does not have any shared config. (child source)

To a Mosaic site, the product of Source B should work the same as all the other products from Source A.  This includes using the shared config of Source A in pages that come from Source B.

This change adds the above feature by allowing a source to apply it's shared config to another source **if they share the same namespace**

## Timeline
- Source A loads it pages and the shared config plugin creates the shared-config
- Source B loads its pages and writes to the global config explaining that it doesn't have any shared config and that it has namespace "whatever"
- The afterUpdate plugin lifecycle triggers for all sources 
- Source A detects that it has the same namespace as Source B
- For each index page in Source B, Source A works out what the closest shared config file should be
- Source A then creates a file in the shared filesystem so that Source B now has a shared config

### DIFF
--- a/packages/plugins/src/SharedConfigPlugin.ts
+++ b/packages/plugins/src/SharedConfigPlugin.ts
@@ -58,7 +58,7 @@ const SharedConfigPlugin: PluginType<SharedConfigPluginPage, SharedConfigPluginO
     if (indexPagesWithSharedConfig.length === 0 && indexPages.length > 0) {
       const rootPath = indexPages[0].fullPath;
       const applyNamespaceSharedConfig = {
-        [`${namespace}-${rootPath}`]: {
+        [`${namespace}~~${rootPath}`]: {
           paths: indexPages.map(indexPage => indexPage.fullPath),
           rootPath,
           namespace
@@ -153,7 +153,7 @@ const SharedConfigPlugin: PluginType<SharedConfigPluginPage, SharedConfigPluginO
     const { applyNamespaceSharedConfig } = globalConfig.data;
 
     if (applyNamespaceSharedConfig === undefined) {
-      // there is no source that exists that has told us it needs to share a parent shared config
+      // there is no source that exists that has told us it needs to share a namespace shared-config
       return;
     }
 
@@ -164,7 +164,7 @@ const SharedConfigPlugin: PluginType<SharedConfigPluginPage, SharedConfigPluginO
       namespace: string;
     }[] = Object.keys(applyNamespaceSharedConfig)
       .filter(key => {
-        const keyNamespace = key.split('-')?.[0];
+        const keyNamespace = key.split('~~')?.[0];
         return keyNamespace === namespace;
       })
       .map(key => applyNamespaceSharedConfig?.[key] || []);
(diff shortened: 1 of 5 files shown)


packages/plugins/src/SharedConfigPlugin.ts
```diff
@@ -43,17 +43,31 @@ export interface SharedConfigPluginOptions {
...
+    if (indexPagesWithSharedConfig.length === 0 && indexPages.length > 0) {
+      const rootPath = indexPages[0].fullPath;
+      const applyNamespaceSharedConfig = {
+        [`${namespace}~~${rootPath}`]: {
+          paths: indexPages.map(indexPage => indexPage.fullPath),
+          rootPath,
+          namespace
+        }
+      };
+      config.setData({ applyNamespaceSharedConfig });
```
DavieReid, line 67: the config object is persisted across plugin lifecycle events.  So this source is writing to the global config int he hope another source that does have a shared config will apply it.

This is picked up in the `afterUpdate` event

packages/plugins/src/SharedConfigPlugin.ts
```diff
@@ -113,24 +127,97 @@ const SharedConfigPlugin: PluginType<SharedConfigPluginPage, SharedConfigPluginO
...
-      );
-      config.setAliases(closestSharedConfig, [sharedConfigFile]);
+    // find all the entries that match the namespace the plugin is running against
+    const namespaceSharedConfigs: {
+      paths: string[];
+      rootPath: string;
+      namespace: string;
+    }[] = Object.keys(applyNamespaceSharedConfig)
+      .filter(key => {
+        const keyNamespace = key.split('~~')?.[0];
```
DavieReid, line 167: There may be more that one source in the namespace that needs a shared config.  I have assumed the sources have different root paths but that might need revisited the more I think about it.  If the rootPath and the namespace are the same for 2 sources, 1 will overwrite the other

joshwooding (approved):

packages/plugins/src/SharedConfigPlugin.ts
```diff
@@ -113,24 +127,97 @@ const SharedConfigPlugin: PluginType<SharedConfigPluginPage, SharedConfigPluginO
...
+        return keyNamespace === namespace;
+      })
+      .map(key => applyNamespaceSharedConfig?.[key] || []);
+
+    for (const namespaceSharedConfig of namespaceSharedConfigs) {
+      if (await mutableFilesystem.promises.exists(namespaceSharedConfig.rootPath)) {
```
DavieReid, line 173: This is to guard against the actual write of the shared config.json that is a copy.  It might trigger afterUpdate again but not 100% sure

packages/plugins/src/SharedConfigPlugin.ts
DavieReid: Updated this to use `crypto.RandomUUID()`

PR #453, opened 2023-08-22
Handle plugin errors

DavieReid:
The primary change included here is that any exceptions raised by plugins during any part of the plugin lifecycle are converted to instances of `PluginError` and tracked by the source that is running the plugins.

This means plugin errors do not cause the source to close.

Plugin authors should be encouraged to throw a`PluginError`.  Should an error occur when processing a particular page, then the full path to the page can be included in the error descriptor.

I have also removed the `saveContent` plugin API.  This was replaced with `workflows` some time ago and should have been removed then.

### Todo

Now that a source has a collection of plugin errors....how are these surfaced?


packages/core/src/worker/Source.worker.ts
```diff
@@ -5,14 +5,22 @@ import fs from 'fs';
...
 
 import FileAccess from '../filesystems/FileAccess.js';
 import MutableVolume from '../filesystems/MutableVolume.js';
 import createConfig from '../helpers/createConfig.js';
 import { bindPluginMethods, bindSerialiser } from '../plugin/index.js';
 import createSourceObservable from './helpers/createSourceObservable.js';
 
+const trackPluginErrors = (errors: PluginErrors, lifecycleMethod: string) => {
+  const errorsBuffer = Buffer.from(JSON.stringify({ errors, lifecycleMethod }));
+  parentPort.postMessage({
```
DavieReid, line 18: Plugin methods that start with a `$` run inside a child process and must communicate with the parent thread by posting a message.

packages/core/src/Source.ts
```diff
@@ -36,6 +37,7 @@ export default class Source {
   #ignorePages: string[];
   #workflows: SourceWorkflow[] = [];
   #schedule: SourceSchedule;
+  #pluginErrors: any = {};
```
DavieReid, line 40: Will update the type of this when we have identified what we want to do with the errors.  They are logged to the console but it would be nice to provide an API that can be used to see the errors

packages/core/src/Source.ts
```diff
@@ -230,7 +235,12 @@ export default class Source {
   }
 
   async restart() {
+    console.log(this.#pluginErrors);
```
joshwooding, line 238: Do you want this? If so should it be console.error instead?

packages/core/src/plugin/pluginRunner.ts
```diff
@@ -20,48 +21,47 @@ export default async function pluginRunner(
...
-      }
-      throw new PluginError(
-        `Plugin '${plugin.modulePath}' threw an exception running \`${lifecycleName}\`. See below:
-  ${e.stack}`
+    } catch (exception) {
+      const pluginName = path.posix.basename(
+        plugin.modulePath,
+        path.posix.extname(plugin.modulePath)
+      );
+      console.log(
```
joshwooding, line 43: Should this be console.error too?

packages/core/src/Source.ts
DavieReid: no...will remove

joshwooding (approved):

packages/core/src/Source.ts
mark-tate: might be worth leaving this till after we merged in Next 13 changes... as right now, it's easy to rebase that branch, when we confine other changes to core and not change any of the UI

mark-tate (approved):

packages/core/src/Source.ts
DavieReid: That's fine...just pushed a change that will allow you to see the plugin errors when using the list sources admin API.  Should be good enough for us to diagnose problems.

packages/core/src/plugin/pluginRunner.ts
DavieReid: DONE

PR #454, opened 2023-08-23
Fix pages that are excluded from the sidebar having no sidebar

joshwooding:


DavieReid:
This plugin has a test suite so would be good to add a test covering the issue this fixes

DavieReid (approved):

PR #456, opened 2023-08-24
Plugin updates

DavieReid:


packages/plugins/src/$AliasPlugin.ts
```diff
@@ -45,21 +50,25 @@ const $AliasPlugin: PluginType<AliasPluginPage> = {
...
+          if (!(await mutableFilesystem.promises.exists(aliasDir))) {
+            await mutableFilesystem.promises.mkdir(aliasDir, {
+              recursive: true
+            });
+          }
+          if (!(await mutableFilesystem.promises.exists(aliasPath))) {
+            await mutableFilesystem.promises.symlink(fullPath, aliasPath);
+          } else {
+            console.warn(
+              `Alias '${aliasPath}' already exists. Is there a duplicate alias, or a page with the same name but different file extensions?`
```
tambo, line 66: Adding some log tagging like [Mosaic][< error category>] would help with monitoring/alerting.

tambo (approved):

packages/plugins/src/$AliasPlugin.ts
DavieReid: Let's create a ticket and document what we want

PR #466, opened 2023-08-31
Inline code block styling

DavieReid:


feliciackh (approved):

JoanaMMoreira (approved):

packages/theme/src/typography/code.css.ts
```diff
@@ -11,13 +12,17 @@ export const code = recipe({
       regular: style([
         {
           fontSize: vars.fontSize.s70,
+          paddingLeft: vars.space.horizontal.x2,
```
origami-z, line 15: inline code with x2 inline padding feels pretty wide to me in a paragraph

DavieReid: Yep, it very much does look like a pill. Im sure the designers will revisit this in the future.

PR #481, opened 2023-09-11
Sidebar helplinks

DavieReid:


packages/store/src/types/sharedConfig.ts
```diff
@@ -1,14 +1,13 @@
...
 import type { FooterSlice } from './footer';
-import type { SidebarSlice } from './sidebar';
 
 export type SharedConfig = {
   /** Footer props */
   footer?: FooterSlice;
   /** Header props */
   header?: AppHeaderSlice;
   /** Sidebar props */
-  sidebar?: SidebarSlice;
```
DavieReid, line 11: This prop is not the same as the data used to generate the sidebar nav

joshwooding (approved):

PR #486, opened 2023-09-15
Remove overrides for Salt Icon color

lilyvc:
This PR removes the overrides for Salt Icon color. The override is stopping the color prop from working on the Salt site, (added here: https://github.com/jpmorganchase/salt-ds/pull/2462).

The Salt icon will now use currentColor by default, so the override shouldn't be needed anymore.


packages/theme/src/salt/salt.css.ts
```diff
@@ -32,14 +32,6 @@ createGlobalTheme('.saltIcon', iconSizeVars, {
   size: '1'
 });
```
mark-tate, line 33: can we remove the icon size as well ?

libertymayc (approved):

DavieReid (approved):

DavieReid (changes requested):
This isnt building...


```
@jpmorganchase/mosaic-theme:build: ✘ [ERROR] [plugin vanilla-extract] Build failed with 1 error:
--
08:56:20.773 | @jpmorganchase/mosaic-theme:build: src/icon/icons.css.ts:5:9: error: No matching export in "src/salt/index.ts" for import "iconSizeVars"

```

PR #487, opened 2023-09-20
Remove Salt icon color and style overrides

lilyvc:
This PR removes the overrides for Salt Icon color and size. 

The color override is stopping the color prop from working on the Salt site, (added here: https://github.com/jpmorganchase/salt-ds/pull/2462). The Salt icon will now use currentColor by default, so the override shouldn't be needed anymore.

The Mosaic Icon now uses number for size to align with Salt.


packages/components/src/Icon/index.tsx
```diff
@@ -1,15 +1,13 @@
...
 
-import styles from './styles.css';
-
 export interface IconProps {
   /** Additional class name for root class override. */
   className?: string;
   /** Name of the icon */
   name: IconNames;
   /** Size of Icon */
-  size?: 'small' | 'medium' | 'large';
```
joshwooding, line 12: Do teams use Icon directly? If so will this cause adoption issues? It it worth supporting both and then mapping to the new values?

lilyvc: I had a look in developer docs and couldn't find any uses of size

lilyvc: I couldn't find any uses of it in the main docs repo

lilyvc: It would be safer to keep this but I don't know if its necessary

DavieReid: I think we should keep it for now.

DavieReid (approved):

PR #490, opened 2023-09-22
489 link editing

DavieReid:


packages/content-editor-plugin/src/components/Toolbar/InsertLink.tsx
```diff
@@ -78,7 +73,7 @@ export const InsertLinkDialog = () => {
         const selection = $getSelection();
         if ($isRangeSelection(selection)) {
           const textContent = selection.getTextContent();
-          setValues(values => ({ ...values, text: textContent }));
+          setValues(currentValues => ({ ...currentValues, text: textContent }));
```
tambo, line 76: currentValues isn't a thing here - should update`values` on line 55 where int with useState

tambo (changes requested):
Needs comment about currentValues addressed or will break

packages/content-editor-plugin/src/components/Toolbar/InsertLink.tsx
DavieReid: This is a functional `setState`.  `currentValues` is the previous state.

See --> https://react.dev/reference/react/useState#setstate

DavieReid: I will rename the variable to make it clearer

DavieReid: Updated

packages/content-editor-plugin/src/components/Toolbar/InsertLink.tsx
```diff
@@ -118,35 +112,29 @@ export const InsertLinkDialog = () => {
...
+          </FormField>
+          <FormField validationStatus={errors?.text ? 'error' : undefined}>
+            <FormFieldLabel>Link Text</FormFieldLabel>
+            <Input value={values?.text} inputProps={{ name: 'text' }} onChange={handleChange} />
+            <FormFieldHelperText>{errors?.text}</FormFieldHelperText>
+          </FormField>
+        </div>
+      </DialogContent>
+      <DialogActions>
+        <ButtonBar>
```
ZarkoPernar, line 131: No need for ButtonBar in actions right now. Although the actions don't have the full functionality of the old button bar. But if this works well then maybe leave it. I haven't tried it myself.

ZarkoPernar (approved):

PR #494, opened 2023-09-27
Breadcrumbs after update

DavieReid:
Followup to https://github.com/jpmorganchase/mosaic/pull/492

The breadcrumbs were still not showing up from a cold start.

This PR introduces a new plugin lifecycle method `shouldUpdateNamespaceSources`.

When a source updates, **other sources that share the same namespace** will re-run `afterUpdate` if there is any plugin that returns true when `shouldUpdateNamespaceSources` is called.

This allows the other source to have a more holistic view of the namespace pages.

### DIFF
--- a/packages/core/src/Source.ts
+++ b/packages/core/src/Source.ts
@@ -221,13 +221,13 @@ export default class Source {
 
     if (foundWorkflows.length === 0) {
       return {
-        error: `[Mosaic] Workflow ${name} not found for ${this.id.description.toString()}`
+        error: `[Mosaic][Source] workflow ${name} not found for ${this.id.description.toString()}`
       };
     }
 
     if (foundWorkflows.length > 1) {
       return {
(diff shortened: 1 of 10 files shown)


packages/core/src/Source.ts
```diff
@@ -130,6 +134,9 @@ export default class Source {
     }
     if (shouldInvokeAfterUpdate === true) {
       this.filesystem.clearCache();
+      this.filesystem.unfreeze();
+      await this.invokeAfterUpdate(sharedFilesystem, globalConfig);
```
DavieReid, line 138: This may be better as another lifecycle method...not sure what to call it but we need a way for a sources plugins to be able to respond to ***another*** source changing

mark-tate (approved):

packages/core/src/Source.ts
DavieReid: Added `shouldUpdateNamespaceSources` lifecycle method

PR #525, opened 2023-11-17
refactor: improve namespace source update plugin lifecycle

DavieReid:


packages/core/src/WorkerSubscription.ts
```diff
@@ -59,10 +59,12 @@ export default class WorkerSubscription {
     } else if (type === 'track') {
       this.#emitter.emit(EVENT.TRACK, { data: JSON.parse(textDecoder.decode(data)) });
     } else if (type === 'init') {
-      this.#emitter.emit(EVENT.START);
       if (data) {
-        // We receive a file buffer for 'init', not an object
-        this.#emitter.emit(EVENT.UPDATE, { data: JSON.parse(textDecoder.decode(data)) });
+        // We receive a file buffer (cached fs) for 'init', not an object
+        this.#emitter.emit(EVENT.START, { data: JSON.parse(textDecoder.decode(data)) });
```
DavieReid, line 64: I think we should always emit a `start` event when the worker thread emits `init`.

When `update` was used it would trigger 2 rounds of running plugins when only 1 was required.

origami-z (approved):

PR #528, opened 2023-11-22
Storybook source david

mark-tate:


lilyvc:
Could this be used on the Salt website, particularly for showcasing component and pattern examples? 

It would be good to centralize our code examples, as they are currently diverging between the website and Storybook. 

The main requirement is the capability to display the code alongside the example.

mark-tate:
> Could this be used on the Salt website, particularly for showcasing component and pattern examples?
> 
> It would be good to centralize our code examples, as they are currently diverging between the website and Storybook.
> 
> The main requirement is the capability to display the code alongside the example.

Yes it can ..
Or any other website where you want to create a page of storybook samples (or groups of storybook stories).
The idea is that you pull into the Mosaic VFS a list of all the stories that match filters (defined by your config).
Each group of stories can be enhanced with additional metadata (what product it relates to, who owns it)

Once the data is in the VFS, you can create pages, which query those stories and build pages from the metadata.
These types of pages generate themselves and are always up-to-date, based on the original Storybook content.
ie. no need to republish the Mosaic page when the stories change

A powerful use-case for Mosaic, is pulling from a data source, normalising and enhancing the data, then using that metadata to generate pages from it's own data.

DavieReid (approved):

PR #542, opened 2023-12-07
Add figma source david

DavieReid:


packages/source-http/src/createHttpSource.ts
```diff
@@ -39,10 +39,11 @@ export interface CreateHttpSourceParams<TResponse, TPage>
  * For use inside *other* sources.
  * Allows a transformer function to be passed directly without the need for dynamic imports.
  *
+ * SourceConfig is also optional here so that scheduling can be ignored.
  */
-export function createHttpSource<TResponse, TPage extends Page>(
+export function createHttpSource<TResponse, TPage = Page>(
```
DavieReid, line 44: This no longer really creates a source because:

1. it can return things that are not pages
2. it can run unscheduled

Another name needed?  `createHttp$`?

packages/source-figma/src/index.ts
```diff
@@ -1,28 +1,29 @@
...
 import createFigmaPage from './transformer.js';
-import { FigmaPage, ProjectFilesResponseJson, ProjectResponseJson } from './types/index.js';
+import type {
+  FigmaPage,
+  FileUrlAndMeta,
+  ProjectFilesResponseJson,
+  ProjectResponseJson
+} from './types/index.js';
 
 const JPM_PATTERN_PREFIX = 'jpmSaltPattern.';
```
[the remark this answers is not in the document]
DavieReid: DONE: this is now a property of `projects` in the mosaic config file

packages/source-http/src/createHttpSource.ts
mark-tate: does a source have to return a page ?
Maybe it's ok calling it a Source ?

DavieReid: I think sources should return pages....but this here can be used to build other sources, that's the difference

PR #560, opened 2024-02-04
Add sitemap component

mark-tate:


DavieReid:
Looks like some docs have been updated so you will need to generate and commit a new snapshot.

DavieReid (approved):

PR #571, opened 2024-02-22
Fix the TOC indentation

joshwooding:


packages/plugins/src/TableOfContentsPlugin.ts
```diff
@@ -43,7 +43,8 @@ const TableOfContentsPlugin: PluginType<TableOfContentsPluginPage, TableOfConten
           .map(validHeading => {
             const text = parsePageHeading(page, validHeading.value);
             return {
-              level: validHeading.depth,
+              depth: validHeading.depth,
+              level: validHeading.depth - minRank,
```
joshwooding, line 47: This was the simplest way to be consistent. Now the level will be relative to the minRank so by default h2 = level 0

DavieReid:
The css needs fixed on the salt site...the padding is too big

DavieReid (approved):

PR #576, opened 2024-02-26
Split storybook urls

mark-tate:


mark-tate:
This created another iteration on naming and alignment across Sources.

Storybook url became 'storiesUrl' and 'storyUrlPrefix'
It generates Pages that have a `link` and a 'contentUrl' meta.
Used that naming for the Figma pages as well, so 'thumbnailUrl' became 'contentUrl'

packages/source-storybook/src/index.ts
```diff
@@ -23,7 +23,8 @@ export const schema = baseSchema.merge(
       .array(
         z.object({
           description: z.string(),
-          url: z.string(),
+          storiesUrl: z.string().url(),
+          storyUrlPrefix: z.string().url(),
```
DavieReid, line 27: Should this be optional?  If not provided `storiesUrl` will be used.

tambo (approved):

packages/source-storybook/src/index.ts
tambo: Agreed. It's probably going to be rarely needed.

mark-tate: done

PR #578, opened 2024-03-01
fix: home page logo should be an svg which when dark mode is selected…

mark-tate:
… will be inverted


packages/site-components/src/AppHeader/AppHeader.tsx
```diff
@@ -50,7 +50,7 @@ export const AppHeader: React.FC<AppHeaderProps> = ({ homeLink, logo, menu = [],
           <Link className={styles.logoContainer} href={homeLink} variant="component">
             {logo && (
               <Logo>
-                <LogoImage src={logo} alt="" />
+                <LogoImage className={styles.logoImage} src={logo} alt="homepage link" />
```
joshwooding, line 53:
```suggestion
                <LogoImage className={styles.logoImage} src={logo} alt="homepage" />
```

tambo (approved):
approved (can adjust alt text in follow up if required)

PR #586, opened 2024-03-14
Fix tag plugin

joshwooding:


packages/plugins/src/$TagPlugin.ts
```diff
@@ -85,7 +85,7 @@ const $TagPlugin: PluginType<TagPluginPage, unknown, TagPluginConfigData> = {
             config.setGlobalRef(
               page.fullPath,
               ref.$$path,
-              `${path.posix.join('/.tags', tag, '**')}#${fragment}`
+              `${path.posix.join('/.tags', tag, '**', '*')}#${fragment}`
```
joshwooding, line 88: pesky globs...

 This was something like `/.tags/in-stock/**{.mdx,.json}` before which isn't valid.

Now its `/.tags/in-stock/**/*{.mdx, .json}`

DavieReid (approved):

packages/plugins/src/$TagPlugin.ts
DavieReid: 👏🏻

PR #602, opened 2024-05-22
refactor: tactical fix to remove `react-pro-sidebar` and replace with…

mark-tate:
… Salt's `NavigationItem`

We will remove this local `NavigationItem` once Salt merges it's own PR.

### DIFF
--- a/packages/site-components/src/NavigationItem/NavigationItem.tsx
+++ b/packages/site-components/src/NavigationItem/NavigationItem.tsx
@@ -1,0 +1,183 @@
+import React, { AriaAttributes, ForwardedRef, MouseEvent, ReactElement } from 'react';
+import { ComponentPropsWithoutRef, forwardRef, MouseEventHandler } from 'react';
+import { makePrefixer } from '@salt-ds/core';
+import { clsx } from 'clsx';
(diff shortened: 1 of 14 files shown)


packages/site-components/src/NavigationItem/NavigationItem.tsx
```diff
@@ -0,0 +1,183 @@
...
+export interface NavigationItemElementProps<T extends HTMLAnchorElement | HTMLButtonElement>
+  extends Pick<AriaAttributes, 'aria-label' | 'aria-expanded' | 'aria-current'> {
+  /**
+   * Item's children.
+   */
+  children: ReactElement;
+  /**
+   * Item's className.
+   */
+  className: string;
```
origami-z, line 23: `?`?

origami-z (approved):
need changeset?

packages/site-components/src/NavigationItem/NavigationItem.tsx
mark-tate: We always pass a className, it's not optional, unlike onClick, which we only pass for parent elements (groups)

PR #631, opened 2024-07-17
Fix site font not loading from next font

origami-z:
Turn off system font to test on site


yarn.lock
```diff
@@ -2737,6 +2737,17 @@
     "@salt-ds/window" "^0.1.1"
     clsx "^2.0.0"
 
+"@salt-ds/core@^1.32.0":
```
joshwooding, line 2740: Multiple versions of core are installed

joshwooding (approved):

PR #639, opened 2024-08-29
Add filter to SearchIndexPlugin

joshwooding:
Not sure what the correct API should be


DavieReid:
What pages are you trying to exclude?  Can they be hidden pages with a `.` in the filename?

joshwooding:
> What pages are you trying to exclude? Can they be hidden pages with a `.` in the filename?

In Salt, we want to not index example pages but we want them to show on the site. Not sure if `.` will help there?

This might be quite custom though, so maybe we'll need a custom Search plugin anyway, if so it can live there for now

DavieReid (approved):

.changeset/cool-ads-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@jpmorganchase/mosaic-plugins': minor
+---
+
+Add a `filter` option to SearchIndexPlugin, which allows you to exclude pages from the search index.
```
mark-tate, line 5: the plugin  already has include in its naming convention, should we call this excludePages (or similar) rather than filter

PR #656, opened 2024-09-07
Add document assets plugin

mark-tate:


packages/plugins/src/DocumentAssetsPlugin.ts
```diff
@@ -0,0 +1,174 @@
...
+function isUrl(assetPath: string): boolean {
+  try {
+    new URL(assetPath);
+    return true;
+  } catch (_err) {}
+  return false;
+}
+
+function isRelativePath(assetPath: string): boolean {
+  const relativePathRegExp = /^(\.{1,2}\/)+/;
```
joshwooding, line 43: Is it worth highlighting paths are expected to be posix pathing?

Or the regex can be updated:

```suggestion
  const relativePathRegExp = /^(\.{1,2}(\/|\\\\))+/;
```

packages/plugins/src/DocumentAssetsPlugin.ts
```diff
@@ -0,0 +1,174 @@
...
+
+function isUrl(assetPath: string): boolean {
+  try {
+    new URL(assetPath);
+    return true;
+  } catch (_err) {}
+  return false;
+}
+
+function isRelativePath(assetPath: string): boolean {
```
joshwooding, line 42: Is there any issue using `!path.isAbsolute()` ?

packages/plugins/src/DocumentAssetsPlugin.ts
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Plugin that finds assets within the Mosaic filesystem and copies them to the configured `outputDir`.
+ * Documents that create relative references to those images, will be re-written to pull the images from the `outputDir`.
+ */
+const DocumentAssetsPlugin: PluginType<Page, DocumentAssetsPluginOptions> = {
+  async afterUpdate(
+    _mutableFileSystem,
+    _helpers,
+    {
+      assetSubDirs = ['images'],
```
joshwooding, line 81: It could be a lot simpler to make this a string and to force globs to be provided e.g. assetDirs: "**/{images,assets}"

packages/plugins/src/DocumentAssetsPlugin.ts
mark-tate: refactored and removed this function

packages/plugins/src/DocumentAssetsPlugin.ts
mark-tate: that would help solve the other comment

packages/plugins/src/DocumentAssetsPlugin.ts
mark-tate: Good shout.. Might make sense if combining sites through scopes and you don't want to copy all the images

joshwooding (approved):

PR #673, opened 2024-10-22
fix bug in cloning of repos, where worktree directory could not be cr…

mark-tate:
…eated

### DIFF
--- a/packages/plugins/src/DocumentAssetsPlugin.ts
+++ b/packages/plugins/src/DocumentAssetsPlugin.ts
@@ -9,6 +9,7 @@ import remarkParse from 'remark-parse';
 import remarkMdx from 'remark-mdx';
 import remarkStringify from 'remark-stringify';
 import { VFile } from 'vfile';
+import PluginError from './utils/PluginError.js';
 
 interface DocumentAssetsPluginOptions {
   /**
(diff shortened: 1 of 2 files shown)


packages/plugins/src/DocumentAssetsPlugin.ts
```diff
@@ -99,19 +97,25 @@ const DocumentAssetsPlugin: PluginType<Page, DocumentAssetsPluginOptions> = {
 
       let globbedImageDirs;
       try {
+        await fsExtra.access(resolvedSrcDir);
```
joshwooding, line 100: `access` is normally discouraged. Could we just rely on glob failing?

joshwooding: > Do not use fs.access() to check for the accessibility of a file before calling fs.open(), fs.readFile(), or fs.writeFile(). Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file is not accessible.

packages/plugins/src/DocumentAssetsPlugin.ts
```diff
@@ -97,22 +97,17 @@ const DocumentAssetsPlugin: PluginType<Page, DocumentAssetsPluginOptions> = {
...
+          console.warn(`Warning: No entries found for ${assetSubDir}. It may not exist.`);
+          continue;
         }
-      }
-
-      if (!globbedImageDirs || globbedImageDirs?.length === 0) {
-        continue;
+      } catch (err) {
+        console.error(`Error globbing ${assetSubDir} in ${srcDir}:`, err);
+        throw err;
```
DavieReid, line 110: Throw a plugin error and it can be tracked and displayed in the admin APIs 

```throw new PluginError(e.message, page.fullPath);```

Adding fullPath of the page makes it easy to see what page has the issue but its optional and might not apply here

DavieReid (approved):

PR #699, opened 2025-02-11
Update storybook link url to include html doc name

tambo:
this will fix links to V7 storybooks which do not work without it.

### DIFF
--- a/packages/source-storybook/src/index.ts
+++ b/packages/source-storybook/src/index.ts
@@ -78,7 +78,7 @@ const transformStorybookPages = (
         title: storyTitle,
         description,
         contentUrl: `${storyUrlPrefix}/iframe.html?id=${id}&viewMode=story&shortcuts=false&singleStory=true`,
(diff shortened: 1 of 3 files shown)


packages/source-storybook/src/index.ts
```diff
@@ -78,7 +78,7 @@ const transformStorybookPages = (
         title: storyTitle,
         description,
         contentUrl: `${storyUrlPrefix}/iframe.html?id=${id}&viewMode=story&shortcuts=false&singleStory=true`,
-        link: `${storyUrlPrefix}?path=/${type}/${id}`
+        link: `${storyUrlPrefix}/index.html?path=/${type}/${id}`
```
DavieReid, line 81: Does this break older storybook versions that were working without this?

tambo: tested against pepper (v8) and markets (v7) and works. I would expect including index.html is more valid than just appending a querystring to a base url, normally.

joshwooding (approved):

PR #707, opened 2025-02-24
Refactor to remove deps

joshwooding:


patches/memfs+3.4.12.patch
joshwooding, line 1: No more patching :)

PR #732, opened 2025-05-30
Fix Tiles and TileLink

origami-z:
Tiles default to 4 is more sensible for existing docs


packages/components/src/Cards.tsx
```diff
@@ -19,8 +19,8 @@ function layoutChildren(children) {
   });
 }
 
-export const Cards: React.FC<CardsProps> = ({ children, ...rest }) => (
-  <GridLayout style={{ margin: '0px' }} {...rest}>
+export const Cards: React.FC<CardsProps> = ({ children, columns = 4, ...rest }) => (
+  <GridLayout style={{ margin: '0px', ...rest.style }} {...rest}>
```
libertymayc, line 23: not passing columns?

origami-z:
Closing for now. Internal site prod issue is fixed for now.

PR #746, opened 2025-09-24
Add last modified check in thumbnail cache

tayyabataimur:


packages/source-figma/src/utils/thumbnailCache.ts
```diff
@@ -1,6 +1,11 @@
-import fs from 'fs';
+import fs, { existsSync, mkdirSync } from 'fs';
```
joshwooding, line 1: NIT: Do we need to mix the default and destructed imports

